PR #38, opened 2024-08-07
Adding support for subpath for model caching

visheshtanksale:


internal/controller/platform/standalone/nimservice.go
```diff
@@ -313,42 +312,43 @@ func (r *NIMServiceReconciler) syncResource(ctx context.Context, obj client.Obje
 }
 
 // getNIMCachePVC returns PVC backing the NIM cache instance
-func (r *NIMServiceReconciler) getNIMCachePVC(ctx context.Context, nimService *appsv1alpha1.NIMService) (string, error) {
+func (r *NIMServiceReconciler) getNIMCachePVC(ctx context.Context, nimService *appsv1alpha1.NIMService) (appsv1alpha1.PersistentVolumeClaim, error) {
```
slu2011, line 315: Can we change the return value to *appsv1alpha1.PersistentVolumeClaim

So that nil can be returned? From the caller point of view a nil check is easier than compare to appsv1alpha1.PersistentVolumeClaim{}

shivamerla: +1, also please add some tests with subpath option both for NIMCache and NIMService

visheshtanksale: Fixed the method and added tests

slu2011 (approved):

PR #40, opened 2024-08-07
Initial commit to enable Pipeline controller

shivamerla:


api/apps/v1alpha1/nimpipeline_types.go
```diff
@@ -23,6 +23,20 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+const (
+	// NIMPipelineConditionReady indicates that the NIM pipeline is ready.
+	NIMPipelineConditionReady = "NIM_PIPELINE_READY"
+	// NIMPipelineConditionFailed indicates that the NIM pipeline has failed.
+	NIMPipelineConditionFailed = "NIM_PIPELINE_FAILED"
+
```
ArangoGutierrez, line 31: let's rm this line

ArangoGutierrez (changes requested):

internal/controller/nimpipeline_controller.go
```diff
@@ -33,30 +37,223 @@ type NIMPipelineReconciler struct {
...
+		return err
+	}
+
+	// Check if the NIMService already exists and create or update as necessary
+	existingService := &appsv1alpha1.NIMService{}
+	err := r.Get(ctx, types.NamespacedName{Name: nimService.Name, Namespace: nimService.Namespace}, existingService)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info("Creating new NIMService", "name", nimService.Name)
+			return r.Create(ctx, nimService)
```
mikemckiernan, line 164: I'm almost certainly making this request at the wrong time and in the wrong place.  Would it be OK to initially set the status to either NotReady or Error when a service (or any of our CRDs) are created until reconciliation can prove they are not?  I ask because I'm an expert at breaking things and I've seen the NIM service status as empty (age too?).  I feel like it'd be nice to say if the status is NotReady for more than 10 minutes, it's time to get the logs from the NIM Operator.  (And also like to avoid typing "NotReady or blank".)

mikemckiernan (commented):
I'm up to my usual tricks.  Thanks.

internal/controller/nimpipeline_controller.go
shivamerla: Totally makes sense, can change it to default to NotReady.

internal/controller/nimpipeline_controller.go
```diff
@@ -33,30 +39,257 @@ type NIMPipelineReconciler struct {
...
+		case appsv1alpha1.NIMServiceStatusPending:
+			// If any service is not ready, mark overall readiness as false
+			allServicesReady = false
+		}
+	}
+
+	// If all services are ready and no failures were detected, set overall state to "Ready"
+	if allServicesReady && overallState != appsv1alpha1.NIMServiceStatusFailed {
+		overallState = appsv1alpha1.NIMServiceStatusReady
+	}
```
visheshtanksale, line 265: Can we copy over the condition of the NIMService to the NIMPipeline, that will give user better picture of whats going on?
https://github.com/NVIDIA/k8s-nim-operator/blob/5a9702b6918fd5dc72b4b0845a3059591c056251/internal/conditions/conditions.go#L111

internal/controller/nimpipeline_controller.go
```diff
@@ -33,30 +39,257 @@ type NIMPipelineReconciler struct {
...
+	// Process each service specification in the pipeline
+	for _, service := range nimPipeline.Spec.Services {
+		enabled := service.Enabled != nil && *service.Enabled
+		enabledServices[service.Name] = enabled
+
+		if !enabled {
+			continue
+		}
+
+		if err := r.reconcileNIMService(ctx, nimPipeline, service); err != nil {
```
visheshtanksale, line 126: Update the condition of NIMPipeline obj here so it does not fail silently.

internal/controller/nimpipeline_controller.go
```diff
@@ -33,30 +39,257 @@ type NIMPipelineReconciler struct {
...
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Clean up disabled NIMServices
+	if err := r.cleanupDisabledNIMs(ctx, nimPipeline, enabledServices); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Update status of NIMServices
```
visheshtanksale, line 136:
```suggestion
	// Update status of NIMPipeline based on the status of related NIMServices
```

internal/controller/nimpipeline_controller.go
```diff
@@ -33,30 +39,257 @@ type NIMPipelineReconciler struct {
...
+	}
+
+	return nil
+}
+
+func (r *NIMPipelineReconciler) cleanupDisabledNIMs(ctx context.Context, nimPipeline *appsv1alpha1.NIMPipeline, enabledServices map[string]bool) error {
+	logger := log.FromContext(ctx)
+
+	serviceList := &appsv1alpha1.NIMServiceList{}
+	if err := r.List(ctx, serviceList, client.InNamespace(nimPipeline.Namespace)); err != nil {
```
visheshtanksale, line 209: Filter the listing of NIM services using owner ref?

internal/controller/nimpipeline_controller.go
```diff
@@ -33,30 +39,257 @@ type NIMPipelineReconciler struct {
...
+			logger.Info("Creating new NIMService", "name", nimService.Name)
+			return r.Create(ctx, nimService)
+		}
+		logger.Error(err, "Failed to get NIMService", "name", nimService.Name)
+		return err
+	}
+
+	// Update the existing NIMService's spec if needed
+	logger.Info("Updating existing NIMService", "name", nimService.Name)
+	existingService.Spec = nimService.Spec
```
visheshtanksale, line 196: Should we compare to check if update is really needed?

internal/controller/nimpipeline_controller.go
shivamerla: updated this.

internal/controller/nimpipeline_controller.go
shivamerla: Changed the logic to continue reconciliation of all services before erroring out. We log the error and capture the status after all services are reconciled. Conditions would need to be added as a separate PR.

internal/controller/nimpipeline_controller.go
shivamerla: We cannot filter based on the owner reference field, but only labels. We are already listing per given namespace, so should be fine. I have added Owner ref check as well below for gathering statuses.

internal/controller/nimpipeline_controller.go
shivamerla: We cannot add same conditiontype from all services, but need to collect a cumulative reason with single type. Will handle condition updates as a separate PR.

internal/controller/nimpipeline_controller.go
shivamerla: yes, good to check the hash before update. Fixed it.

visheshtanksale (approved):
LGTM

PR #47, opened 2024-08-12
chore: update the status value strings of nim cache and nim services

slu2011:


mikemckiernan (commented):
Is that really all there is to it?  I chickened out because I noticed some more complicated work in https://github.com/NVIDIA/k8s-nim-operator/blob/main/internal/conditions/conditions.go.

slu2011:
> Is that really all there is to it? I chickened out because I noticed some more complicated work in https://github.com/NVIDIA/k8s-nim-operator/blob/main/internal/conditions/conditions.go.

I believe conditions.go is all good (status values consistent with k8s api). Just updating the status value strings of nimcache / nimservices should be enough to fix the issue.

slu2011:
> > Is that really all there is to it? I chickened out because I noticed some more complicated work in https://github.com/NVIDIA/k8s-nim-operator/blob/main/internal/conditions/conditions.go.
> 
> I believe conditions.go is all good (status values consistent with k8s api). Just updating the status value strings of nimcache / nimservices should be enough to fix the issue.

visheshtanksale (approved):

mikemckiernan:
> > Is that really all there is to it? I chickened out because I noticed some more complicated work in https://github.com/NVIDIA/k8s-nim-operator/blob/main/internal/conditions/conditions.go.
> 
> I believe conditions.go is all good (status values consistent with k8s api). Just updating the status value strings of nimcache / nimservices should be enough to fix the issue.

Delightful!  So sorry that I was cowardly.  :)

PR #48, opened 2024-08-12
Adding support for setting userID and groupID

visheshtanksale:


api/apps/v1alpha1/nimcache_types.go
```diff
@@ -40,6 +41,8 @@ type NIMCacheSpec struct {
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
 	// NodeSelectors are the node selector labels to schedule the caching job.
 	NodeSelectors map[string]string `json:"gpuSelectors,omitempty"`
+	UserId        *int64            `json:"userId,omitempty"`
```
slu2011, line 44: Nit : change to UserID / groupID -- Not sure why not reported but some golint rules may complain; same for the getter function names

slu2011 (approved):

api/apps/v1alpha1/nimcache_types.go
visheshtanksale: Updated it

slu2011 (approved):

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -239,6 +242,22 @@ func (n *NIMCache) GetPVCName(pvc PersistentVolumeClaim) string {
 	return pvcName
 }
 
+// GetUserID returns user ID. Returns default value if not set on NimCache object.
+func (n *NIMCache) GetUserID() *int64 {
+	if n.Spec.UserID == nil {
+		return ptr.To[int64](1000)
```
elezar, line 248: Should these be constants?

shivamerla: +1

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -480,6 +482,18 @@ func (n *NIMService) GetServicePort() int32 {
 	return n.Spec.Expose.Service.OpenAIPort
 }
 
+// GetUserID returns the user ID for the NIMService deployment
+func (n *NIMService) GetUserID() *int64 {
```
shivamerla, line 486: We should return same defaults here too when not explicitly set, that way files in the PVC are accessible for the NIMService Pod.

PR #51, opened 2024-08-14
Set NIMService Status.State with public values

ArangoGutierrez:
This change is needed to allow a programmatic check of the NIMService `Status.State`
Conditions should ALWAYS be public values, so third-party customers or our own test suites can perform assertions over the CR objects.


mikemckiernan (approved):
I like the looks of these changes.

slu2011 (approved):

PR #53, opened 2024-08-15
Fix check for external PVC

shivamerla:
fix model store selection using below order
1. Use NIMCache instance if specified for NIMService
2. Create a new PVC if creation is enabled
3. Use an external/existing PVC if the name is provided


mikemckiernan (approved):
I like it! Thank you!

mkhaas (approved):

shivamerla:
> Can we please add a check that either PVC or nimcache can be specified in nimservice? thanks

that is already present, if none are specified, we will error out.

ArangoGutierrez (approved):

shivamerla:
> thank you, I meant
> 
> > > Can we please add a check that either PVC or nimcache can be specified in nimservice? thanks
> > 
> > 
> > that is already present, if none are specified, we will error out.
> 
> thank you, I meant that customer can only specify one .. either nimcache or PVC not both

If both are provided, then nimcache will have the priority. Are you suggesting to error out than applying the priority?

PR #54, opened 2024-08-16
Add role/rolebinding to let nimcache and nimservice instances to use "nonroot" scc

shivamerla:
nonroot SCC is created by default in OCP and allow pods to run with any-uid other than root user. Add necessary roles and rolebinding to allow access for nimcache and nimservice pods.


internal/controller/nimcache_controller.go
```diff
@@ -706,7 +889,10 @@ func constructJob(nimCache *appsv1alpha1.NIMCache) (*batchv1.Job, error) {
 							},
 						},
 					},
-					ImagePullSecrets: []corev1.LocalObjectReference{},
+					ImagePullSecrets:   []corev1.LocalObjectReference{},
+					ServiceAccountName: nimCache.GetName(),
```
slu2011, line 893: Can you double check if service account name should be set as the nim cache name

shivamerla: That is a standard practice to have unique SA per deployment. Are you recommending to create one per namespace?

slu2011: For now it should be simpler if all nim cache job in one ns share the same sa/role, instead of per nim cache instance.

shivamerla: done, updated to create a common role/rolebinding/sa per namespace

slu2011 (approved):

PR #62, opened 2024-08-16
Add default service port for NIM as 8000

shivamerla:


visheshtanksale (approved):
LGTM.
nit: should we just name that field has `port` instead of `OpenAIPort`

shivamerla:
> LGTM. nit: should we just name that field has `port` instead of `OpenAIPort`
sounds good, once NIM supports different ports for health checks, we can change them to healthPort and openAIPort.

PR #63, opened 2024-08-16
Error out when no profile selection is made and no GPUs are assigned …

shivamerla:
…for auto-selection


visheshtanksale (approved):
LGTM!

PR #72, opened 2024-08-18
Add E2E tests

ArangoGutierrez:


shivamerla:
@ArangoGutierrez why are these merged without review?

PR #80, opened 2024-08-19
Install dependencies before e2e during GHA

ArangoGutierrez:


ArangoGutierrez:
@shivamerla My bad, since it worked on my local dev, I forgot to also install `ginkgo` bin during the action run

shivamerla (approved):

PR #82, opened 2024-08-19
Adding support for readonly model PVCs

visheshtanksale:


internal/controller/platform/standalone/nimservice_test.go
```diff
@@ -114,7 +114,8 @@ var _ = Describe("NIMServiceReconciler for a standalone platform", func() {
 					},
 				},
 				NIMCache: appsv1alpha1.NIMCacheVolSpec{
-					Name: "test-nimcache",
+					Name:     "test-nimcache",
+					ReadOnly: true,
```
slu2011, line 118: How about keep the old testcase and assert readonly is false (default value) then add the new testcase such that readOnly is set to true?

slu2011 (dismissed):

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -132,6 +132,7 @@ type PersistentVolumeClaim struct {
 	// VolumeAccessMode is the volume access mode of the PVC
 	VolumeAccessMode corev1.PersistentVolumeAccessMode `json:"volumeAccessMode,omitempty"`
 	SubPath          string                            `json:"subPath,omitempty"`
+	ReadOnly         bool                              `json:"readOnly,omitempty"`
```
shivamerla, line 135: `ReadOnly` should not be part of NIMCache volume spec as we always need it to be rw for caching, instead we need to add it under `NIMService`.

visheshtanksale: I have added `ReadOnly` to `NIMCacheVolSpec`. 
Its here on the PersistenceVolume to support use case in which user has created and populated the PVC without using NIMCache.

internal/controller/platform/standalone/nimservice_test.go
visheshtanksale: Added more tests

api/apps/v1alpha1/nimcache_types.go
visheshtanksale: Fixed after update to the CRD

shivamerla (approved):
lgtm

PR #83, opened 2024-08-19
Add GitLab Pipelines for Image scanning

ArangoGutierrez:


ArangoGutierrez:
cc @tariq1890 @cdesiniotis Please help me with an eye here

.common-ci.yml
```diff
@@ -0,0 +1,211 @@
...
+    PLATFORM: linux/amd64
+
+.platform-arm64:
+  variables:
+    PLATFORM: linux/arm64
+
+# Make buildx available as a docker CLI plugin
+.buildx-setup:
+  before_script:
+    -  export BUILDX_VERSION=v0.6.3
```
cdesiniotis, line 88: nit: this is a pretty old version of buildx. Can we at least upgrade it to v0.15.1 which we use in the gpu-operator? https://github.com/NVIDIA/gpu-operator/blob/main/.common-ci.yml#L74. I see we already disable attaching provenance and sbom attestations to the image, so we can upgrade safely without having this additional metadata added to the resulting images.

.common-ci.yml
```diff
@@ -0,0 +1,211 @@
...
+    - .dist-distroless
+    - .platform-arm64
+  needs:
+    - image-distroless
+    - scan-distroless-amd64
+
+# Download the regctl binary for use in the release steps
+.regctl-setup:
+  before_script:
+    - export REGCTL_VERSION=v0.3.10
```
cdesiniotis, line 142: nit: similar to my comment on buildx, can we upgrade the regctl version to at least v0.7.1? https://github.com/NVIDIA/gpu-operator/blob/main/.common-ci.yml#L153

.nvidia-ci.yml
```diff
@@ -0,0 +1,172 @@
...
+
+sign:ngc-short-tag:
+  extends:
+    - .sign:ngc
+  needs:
+    - release:ngc-distroless
+  variables:
+    IMAGE_TAG: "${CI_COMMIT_TAG}"
+
+sign:ngc-distroless:
```
cdesiniotis, line 167: Question: what will the resulting tag look like? Will it be `${CI_COMMIT_TAG}-distroless`? Do we even need to publish this tag? Or is the short tag sufficient?

ArangoGutierrez: Just short tag, I left the distrolless there as a reference, but the tag won't have it attached

ArangoGutierrez:
Thanks for all the comments, ready for re-review  @cdesiniotis

slu2011 (approved):

cdesiniotis (approved):

PR #87, opened 2024-08-19
Fixing NIMCache creation when profiles are specified

visheshtanksale:


internal/controller/nimcache_controller.go
```diff
@@ -495,56 +495,57 @@ func getSelectedProfiles(nimCache *appsv1alpha1.NIMCache) ([]string, error) {
...
-		if err := controllerutil.SetControllerReference(nimCache, pod, r.GetScheme()); err != nil {
-			return false, err
-		}
-		err := r.createPod(ctx, pod)
-		if err != nil {
-			logger.Error(err, "failed to create", "pod", pod.Name)
-			return false, err
-		}
+	// Create a temporary pod for parsing model manifest
+	pod := constructPodSpec(nimCache)
```
shivamerla, line 502: we should not run the pod on every reconciliation, but only when the configmap is not already created.

PR #88, opened 2024-08-20
Extract NIM model manifest in all cases for model caching

shivamerla:


internal/controller/nimcache_controller.go
```diff
@@ -495,73 +495,112 @@ func getSelectedProfiles(nimCache *appsv1alpha1.NIMCache) ([]string, error) {
...
-			logger.Error(err, "Failed to create model manifest config map")
-			return false, err
-		}
+	if existingPod.Status.Phase != corev1.PodRunning {
+		// requeue request with delay until the pod is ready
+		return true, nil
+	}
 
+	// Extract manifest file
+	output, err := r.getPodLogs(ctx, existingPod)
```
slu2011, line 545: For model manifests -- do we have plan to add NGC api (or any other web apis) to query model metadata?

slu2011: pull an image and cat a file inside looks very hacky......

shivamerla: that is not planned in the near term. The manifest is embedded in the NIM container itself. So we need to run the container to get that (either by `cp` or logging)

PR #94, opened 2024-08-21
Update samples

shivamerla:


config/samples/apps_v1alpha1_nimcache.yaml
```diff
@@ -6,4 +6,25 @@ metadata:
...
+        profiles: []
+        autoDetect: true
+        precision: "fp8"
+        engine: "tensorrt_llm"
+        qosProfile: "throughput"
+        gpus:
+          product: "l40s"
+          ids:
+            - "26b5"
+        tensorParallelism: "1"
```
mikemckiernan, line 24: FWIW, Meenakshi flagged me for having `1` in this field.  Some concern that readers might think it's necessary to specify `1` rather than omit for `1`.  Clearly up to you.

bundle/manifests/k8s-nim-operator.clusterserviceversion.yaml
```diff
@@ -73,13 +61,18 @@ metadata:
...
+                "name": "meta-llama3-8b-instruct",
+                "profile": ""
+              }
             },
             "scale": {
-              "minReplicas": 1
+              "enabled": false,
+              "hpa": {
+                "minReplicas": 1,
+                "maxReplicas": 10
```
mikemckiernan, line 74: This helps me see that I had a misunderstanding.  I thought the `hpa` field would be identical to some common Kubernetes API object.  TY.

mikemckiernan (approved):
TY, except for me telling lies about the `hpa` field, I'm up to date with what you have here.  I appreciate it.

config/samples/apps_v1alpha1_nimcache.yaml
shivamerla: We still need users to specify this for filtering. Otherwise we will end up pulling all models for `l40s` in this example those that will need `4` and `8` GPUs as well. This just makes it explicit on what user needs us to cache.

visheshtanksale (approved):
LGTM

PR #95, opened 2024-08-21
Update nimservice status when deployment fails

visheshtanksale:


shivamerla (commented):
lgtm, can we add some tests with different statuses and conditions.

slu2011 (dismissed):

internal/controller/platform/standalone/nimservice_test.go
```diff
@@ -406,3 +407,166 @@ var _ = Describe("NIMServiceReconciler for a standalone platform", func() {
 		})
 	})
 })
+
+var _ = Describe("isDeploymentReady for setting status on NIMService", func() {
```
shivamerla, line 411: Can we add these as a new context under existing Describe spec?  That way we don't have to initialize all these again. The Context can be "When creating the NIMService deployment" and Tests can be "It should set the status/conditions correctly when progress deadline exceeded" etc.

visheshtanksale:
> lgtm, can we add some tests with different statuses and conditions.

@shivamerla Added unit tests

internal/controller/platform/standalone/nimservice_test.go
visheshtanksale: Good point. Fixed it.

shivamerla (approved):
LGTM! Thanks @visheshtanksale

PR #97, opened 2024-08-22
Requeue requests to fetch pod logs for extracting NIM manifest

shivamerla:
* A delay is required to extract NIM pod logs for the manifest, otherwise lead to empty manifest
* Fix requeue logic to queue with delay
* Tests for parsing embedding/reranking manifests


visheshtanksale (approved):
LGTM!!

PR #98, opened 2024-08-22
Update profile selection for caching to support non-llm manifests

shivamerla:


internal/nimparser/nimparser.go
```diff
@@ -127,75 +127,92 @@ func MatchProfiles(modelSpec appsv1alpha1.ModelSpec, manifest NIMManifest, disco
 	var selectedProfiles []string
 
 	for hash, profile := range manifest {
-		if modelSpec.Precision != "" && profile.Tags["precision"] != modelSpec.Precision {
-			// continue with the next profile in the manifest
+		// Check precision, tensor parallelism, and QoS profile
```
slu2011, line 130: Hopefully we can find a place where nim team defines the keys -- if there are any updates we will know

internal/nimparser/nimparser.go
```diff
@@ -127,75 +127,92 @@ func MatchProfiles(modelSpec appsv1alpha1.ModelSpec, manifest NIMManifest, disco
...
-		if modelSpec.Engine != "" && profile.Tags["llm_engine"] != modelSpec.Engine {
-			// continue with the next profile in the manifest
+
+		// Determine backend type
+		backend := profile.Tags["llm_engine"]
+		if backend == "" {
+			backend = profile.Tags["backend"]
+		}
+
+		if modelSpec.Engine != "" && !strings.Contains(backend, strings.TrimSuffix(modelSpec.Engine, "_llm")) {
```
slu2011, line 151: minor, but can we use == instead of Contains?

slu2011 (approved):

internal/nimparser/nimparser.go
shivamerla: @slu2011 currently we have to rely on NIM repository.

internal/nimparser/nimparser.go
shivamerla: @slu2011 i have used substring match just in case if the backend value turns out be something like `tensorrt_xxx` in the profile. Basically all these are optimized profiles we can filter.

PR #106, opened 2024-08-27
Updating rolling update strategy for nimservice deployment

visheshtanksale:


shivamerla (approved):

shivamerla:
As a follow up we need to expose these through NIMService API (that will work with both a Deployment and a StatefulSet)

mkhaas (commented):
lgtm. Updates will be slow, and one at a time, and needs no new resources.

PR #107, opened 2024-08-27
Update logic to match GPUs only for optimized model selection

shivamerla:


mikemckiernan:
I know you didn't do this for giggles...what happens if someone specifies `engine: vllm`?  Or is the goal to remove that too?  (Not urgent, I'll catch on eventually.)

internal/nimparser/nimparser.go
```diff
@@ -166,6 +177,10 @@ func MatchProfiles(modelSpec appsv1alpha1.ModelSpec, manifest NIMManifest, disco
 	return selectedProfiles, nil
 }
 
+func isOptimizedEngine(engine string) bool {
+	return engine != "" && strings.Contains(strings.ToLower(engine), BackendTypeTensorRT)
```
slu2011, line 181: Are we going to have additional backend types in the future? If yes please add todo....

internal/nimparser/nimparser.go
```diff
@@ -152,11 +157,17 @@ func MatchProfiles(modelSpec appsv1alpha1.ModelSpec, manifest NIMManifest, disco
 			continue
 		}
 
-		// GPU matching logic
-		if len(modelSpec.GPUs) > 0 || len(discoveredGPUs) > 0 {
-			if !matchGPUProfile(modelSpec, profile, discoveredGPUs) {
+		// Perform GPU match only when optimized engine is selected or GPU filters are provided
+		if isOptimizedEngine(modelSpec.Engine) || len(modelSpec.GPUs) > 0 {
```
slu2011, line 161: A model may not use tensorrt engine while still requires gpu, is that right? Seems that for that case there would be no match

shivamerla: No, this case is to still filter optimized profiles, when GPU options are provided, but not the engine type itself in the model spec.

internal/nimparser/nimparser.go
shivamerla: I am not sure on that, we can add if NVIDIA builds for other backends.

shivamerla:
> I know you didn't do this for giggles...what happens if someone specifies `engine: vllm`? Or is the goal to remove that too? (Not urgent, I'll catch on eventually.)

If it is explicitly specified, we match with only those backends/engines. If not, we assume "tensorrt" when GPU/ID filtering is provided, as that only applies to optimized models.

slu2011 (approved):

PR #109, opened 2024-08-27
Updating nimcache crd to auto detect

visheshtanksale:


slu2011 (approved):

shivamerla (dismissed):
LGTM

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -71,8 +71,6 @@ type NGCSource struct {
 type ModelSpec struct {
 	// Profiles are the specific model profiles to cache
```
shivamerla, line 72: We should make this comment more explicit along the lines of "Profiles indicate specific NIM profiles to cache. When these are provided, rest of the model parameters for profile selection are ignored"

mikemckiernan:
Is the goal to select model profiles automatically, by default, unless the model profiles are specified?

shivamerla:
> Is the goal to select model profiles automatically, by default, unless the model profiles are specified?

Yes, we discussed this, the default is to apply profile selection with given parameters, when these are selected explicitly. The `autoDetect` field was redundant in this case. This also means, by default the operator will detect GPUs and apply filtering when explicit GPUs are not provided for the optimized models.

api/apps/v1alpha1/nimcache_types.go
visheshtanksale: Fixed

shivamerla (dismissed):
lgtm, thanks

shivamerla (approved):

PR #111, opened 2024-08-30
Updating NIMService update logic

visheshtanksale:


internal/controller/platform/standalone/nimservice.go
```diff
@@ -290,7 +290,7 @@ func (r *NIMServiceReconciler) isDeploymentReady(ctx context.Context, namespaced
 
 	cond := getDeploymentCondition(deployment.Status, appsv1.DeploymentProgressing)
 	if cond != nil && cond.Reason == "ProgressDeadlineExceeded" {
```
shivamerla, line 292: Should we even have this check?

internal/conditions/conditions.go
```diff
@@ -122,7 +123,21 @@ func (u *updater) SetConditionsFailed(ctx context.Context, cr *appsv1alpha1.NIMS
...
+
+func (u *updater) updateNIMServiceStatus(ctx context.Context, cr *appsv1alpha1.NIMService) error {
+
+	obj := &appsv1alpha1.NIMService{}
+	errGet := u.client.Get(ctx, types.NamespacedName{Name: cr.Name, Namespace: cr.GetNamespace()}, obj)
+	if errGet != nil {
+		return errGet
+	}
+	obj.Status = cr.Status
+	if err := u.client.Status().Update(ctx, obj); err != nil {
```
slu2011, line 137: Can we use Patch() operation instead of Get then Update? It should be more efficient

internal/controller/platform/standalone/nimservice.go
visheshtanksale: The function is used to get the deployment status. I feel the more the details we have on NIMservice we have the better it is.

internal/conditions/conditions.go
visheshtanksale: Since the status field has list of conditions, I am assuming I will have to use MergePatch. I will need the original object to create the merge patch.

shivamerla: agree, patch or server side apply is more resilient for this issue. Also feel free to update this logic across all types as a separate PR.

shivamerla (approved):
lgtm

PR #121, opened 2024-09-03
Add initial Helm chart generator script

ArangoGutierrez:


visheshtanksale (approved):
LGTM!! 
Can we also add a make target?

ArangoGutierrez:
> LGTM!! Can we also add a make target?

Done

hack/package-helm-charts.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -e
```
shivamerla, line 1: Both of these seem redundant, can we just use the make target from the pipeline as well.

ArangoGutierrez: makes sense

ArangoGutierrez: done

Makefile
```diff
@@ -145,6 +145,11 @@ build-installer: manifests generate kustomize ## Generate a consolidated YAML wi
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	$(KUSTOMIZE) build config/default > dist/install.yaml
 
+# Short version of hack/package-helm-charts.sh
+.PHONY: helm-charts
+helm-charts:
```
shivamerla, line 150: @ArangoGutierrez lets add all CI changes required for Helm packaging together in one MR. In this case its not clear which image will be used with this target as that will not match with `VERSION`.

shivamerla: sounds good.

shivamerla (approved):
lgtm

PR #131, opened 2024-09-09
Enforce minimal security context as NIM operator doesn't need elevated privileges

shivamerla:


ArangoGutierrez (changes requested):
Why hardcoding if we can template via helm values? we can enforce from the values, and if the user decides to edit for any specific reason, they should be able to.

shivamerla:
> Why hardcoding if we can template via helm values? we can enforce from the values, and if the user decides to edit for any specific reason, they should be able to.

I would agree these for operands as elevated permissions might be required, but for the operator we can limit these as i see no reason to allow elevated privileges. Also less clutter with helm values. Same applies to CSV file for OLM b.t.w.

ArangoGutierrez (approved):

PR #134, opened 2024-09-10
Automatically inject GPU resources and node selector for NIM deployment 

shivamerla:


internal/nfdutil/nfdutil.go
```diff
@@ -0,0 +1,81 @@
+/*
+Copyright 2024.
```
ArangoGutierrez, line 2:
```suggestion
 * Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```

ArangoGutierrez (changes requested):

internal/controller/platform/standalone/nimservice.go
```diff
@@ -139,89 +146,175 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
...
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *NIMServiceReconciler) setupPVC(ctx context.Context, nimService *appsv1alpha1.NIMService) (*appsv1alpha1.PersistentVolumeClaim, error) {
+	logger := log.FromContext(ctx)
+
+	// Initialize variable for the PVC
```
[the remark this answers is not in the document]
shivamerla: The refactoring was necessary as we had to customize deployment params for multiple reasons now. (PVC, node selector, resources etc)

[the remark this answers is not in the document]
shivamerla: sounds good, let me refactor this into TP change and one for node selector itself.

internal/nfdutil/nfdutil.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	// Check if the NodeFeatureRule CRD exists
+	crdExists, err := checkCRDExists(ctx, k8sClient, "nodefeaturerules.nfd.k8s.io")
+	if err != nil {
+		return false, err
+	}
+
+	if !crdExists {
+		return false, nil
```
slu2011, line 45: return a new error saying that NFD CRD does not exist instead of nil?

internal/controller/platform/standalone/nimservice.go
```diff
@@ -139,89 +146,175 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
...
+			deploymentParams.Resources = &corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{},
+				Limits:   corev1.ResourceList{},
+			}
+		}
+
+		// TODO: Make the resource name configurable
+		const gpuResourceName = corev1.ResourceName("nvidia.com/gpu")
+
+		deploymentParams.Resources.Requests[corev1.ResourceName(gpuResourceName)] = gpuQuantity
```
slu2011, line 292: Can the customer set a gpu resource request / limit in the nim service spec? (I believe yes) and this logic may overwrite the value, is it correct? Also, do we have unit test coverage for this.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -139,89 +146,175 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
...
+		deploymentParams.Resources.Limits[corev1.ResourceName(gpuResourceName)] = gpuQuantity
 	}
 
+	// Apply node selector based on the device ID from the profile
+	// TODO: handle OCP specific logic for NFD API group
+	deviceID, err := r.getDeviceIDByProfile(ctx, profile)
```
slu2011, line 298: Wondering if a model profile can be run with multiple possible gpu types, instead of one type

internal/controller/platform/standalone/nimservice.go
```diff
@@ -372,6 +465,55 @@ func (r *NIMServiceReconciler) getNIMCachePVC(ctx context.Context, nimService *a
...
+	if tp, exists := profile.Config["tp"]; exists {
+		tensorParallelism = tp
+	}
+
+	return tensorParallelism, nil
+}
+
+func (r *NIMServiceReconciler) getDeviceIDByProfile(ctx context.Context, profile *appsv1alpha1.NIMProfile) (string, error) {
+	deviceID := ""
+	if device, exists := profile.Config["gpu_device"]; exists {
```
slu2011, line 479: from the model point of view, the same (model, profile) should be able to run on multiple types of gpus. But here the logic seem to indicate that it is always the 1:1 mapping. Does nim enforce the latter?

shivamerla: @slu2011 models are optimized for specific GPU SKUs, they cannot be run on other types.

internal/controller/platform/standalone/nimservice.go
shivamerla: Yes, if a profile is selected, then we populate the number of GPUs and schedule them to matching GPU type. For generic profile selection, they can specify resources to be consumed.

internal/nfdutil/nfdutil.go
shivamerla: the intent of this function is to check if a specific NodeFeatureRule CR exists, so if the CRD itself is missing we can assume CR is not present as well. Hence error is not thrown in this case.

internal/controller/platform/standalone/nimservice.go
slu2011: How about different mig profiles from the same physical gpu? like 1g.10gb vs  2g.20gb. Does nim provide different model even for different mig profiles, or it would be the same?

shivamerla: @slu2011 currently NIMs need a full GPU, they don't support MIG.

shivamerla:
closing in favor of https://github.com/NVIDIA/k8s-nim-operator/pull/142

PR #142, opened 2024-09-16
Auto assign GPUs for NIM deployments when a cached optimized profile is specified

shivamerla:
Optimized NIM profile config contains a field called tensorParallelism which indicates required number of GPUs. Use that to auto-assign GPU resources for NIM deployments.

### DIFF
--- a/internal/controller/platform/standalone/nimservice.go
+++ b/internal/controller/platform/standalone/nimservice.go
@@ -23,6 +23,7 @@ import (
 	appsv1alpha1 "github.com/NVIDIA/k8s-nim-operator/api/apps/v1alpha1"
 	"github.com/NVIDIA/k8s-nim-operator/internal/conditions"
 	"github.com/NVIDIA/k8s-nim-operator/internal/render"
+	rendertypes "github.com/NVIDIA/k8s-nim-operator/internal/render/types"
 	"github.com/NVIDIA/k8s-nim-operator/internal/shared"
 	"github.com/NVIDIA/k8s-nim-operator/internal/utils"
 	"github.com/go-logr/logr"
@@ -33,6 +34,7 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apiResource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


internal/controller/platform/standalone/nimservice.go
```diff
@@ -413,3 +427,82 @@ func (r *NIMServiceReconciler) reconcilePVC(ctx context.Context, nimService *app
...
+			}
+		}
+
+		// TODO: Make the resource name configurable
+		const gpuResourceName = corev1.ResourceName("nvidia.com/gpu")
+
+		deploymentParams.Resources.Requests[corev1.ResourceName(gpuResourceName)] = gpuQuantity
+		deploymentParams.Resources.Limits[corev1.ResourceName(gpuResourceName)] = gpuQuantity
+	}
+
```
slu2011, line 506: How about add some logging when tensorParallelism is empty?

internal/controller/platform/standalone/nimservice.go
```diff
@@ -413,3 +427,82 @@ func (r *NIMServiceReconciler) reconcilePVC(ctx context.Context, nimService *app
...
+		logger.Error(err, "Failed to retrieve tensorParallelism")
+		return err
+	}
+
+	// Check if tensorParallelism is defined in the profile, and automatically assign GPU resources.
+	// Note: This will override any manual GPU assignments made by the user.
+	// The number of GPUs for an optimized profile is fixed.
+	// Allocating more GPUs than required may result in underutilization,
+	// while allocating fewer GPUs will likely cause failures.
+	if tensorParallelism != "" {
```
slu2011, line 486: if tensorParallelism is empty, then the user specified gpu resource count would be used? Or should it default to 1 gpu?

internal/controller/platform/standalone/nimservice.go
shivamerla: done.

internal/controller/platform/standalone/nimservice.go
shivamerla: Good point. I have updated the logic a bit to honor the request from the user. If none are specified, then auto-assignment is done. This is all when a NIM profile is selected.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -459,50 +459,82 @@ func (r *NIMServiceReconciler) getNIMCacheProfile(ctx context.Context, nimServic
 	return nil, nil
 }
 
+// getTensorParallelismByProfile returns the value of tensor parallelism parameter in the given NIM profile
 func (r *NIMServiceReconciler) getTensorParallelismByProfile(ctx context.Context, profile *appsv1alpha1.NIMProfile) (string, error) {
+	// List of possible keys for tensor parallelism
+	possibleKeys := []string{"tensorParallelism", "tp"}
```
slu2011, line 465: :)

slu2011 (approved):

internal/controller/platform/standalone/nimservice.go
shivamerla: we never know :)

visheshtanksale (approved):
LGTM

PR #154, opened 2024-09-25
Update PVC name in the spec with the once created by the operator

shivamerla:


mikemckiernan:
@shivamerla, just checking...so this update means applying a service manifest like the following will create the PVC (essentially skipping the cache step)?

```
  storage:
    pvc:
      create: true
      storageClass: <storage-class-name>
      size: 10Gi
      volumeAccessMode: ReadWriteMany
```

shivamerla:
> @shivamerla, just checking...so this update means applying a service manifest like the following will create the PVC (essentially skipping the cache step)?
> 
> ```
>   storage:
>     pvc:
>       create: true
>       storageClass: <storage-class-name>
>       size: 10Gi
>       volumeAccessMode: ReadWriteMany
> ```

Yes, that was the behavior prior to this MR. This MR fixes as case where pod bring up was failing without an explicit "pvc.name" in the spec even though a new PVC is created by the operator.

visheshtanksale (approved):
LGTM!!

PR #176, opened 2024-10-03
Adding container name to read NIM Cache logs

visheshtanksale:


shivamerla:
LGTM, there are other places where we assume main container index as 0 in this file, are you planning to handle those separately?

visheshtanksale:
> LGTM, there are other places where we assume main container index as 0 in this file, are you planning to handle those separately?

@shivamerla We are using the main container on the job spec and not pod. So I think we are safe.

mkhaas:
@visheshtanksale - will the NIMService not have a similar issue when sidecar is injected? thanks

shivamerla (approved):

PR #188, opened 2024-10-17
Updating NIMService CRD for multinode inferencing

visheshtanksale:


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -51,6 +51,20 @@ const (
 	NIMServiceStatusFailed = "Failed"
 )
 
+// MultiNodeSpec defines the parameters to render a multi-node NIMService deployment.
+type MultiNodeSpec struct {
+	ClusterStartTimeout int             `json:"clusterStartTimeout,omitempty"`
+	Enabled             *bool           `json:"enabled,omitempty"`
+	Workers             int             `json:"workers,omitempty"`
+	GpusPerNode         int             `json:"gpusPerNode,omitempty"`
```
slu2011, line 59: In the leaderworkerset, does the customer specify the # of GPUs needed for each worker / leader? If the customer specifies that value, GpusPerNode field would become redundant;

Or, do we assume that each worker would take the whole node and would fill GpusPernode to the worker template?

PR #189, opened 2024-10-19
chore: consolidate condition updater for nimservice / guardrail

slu2011:
Unify the condition updater for nimservice / nemoguardrail. Next step is to add a common app reconciler / renderer


shivamerla (approved):
LGTM

PR #196, opened 2024-10-24
Add common metadata required for RedHat certification

shivamerla:


bundle/manifests/k8s-nim-operator.clusterserviceversion.yaml
```diff
@@ -3,7 +3,19 @@ kind: ClusterServiceVersion
 metadata:
   name: nim-operator-certified.v0.1.0
   namespace: placeholder
+  labels:
+    operatorframework.io/arch.amd64: supported
```
slu2011, line 7: For longer term, at one time we need to add arm64 support for grace / blackwell? I believe gpu operator already does that?

slu2011 (approved):

bundle/manifests/k8s-nim-operator.clusterserviceversion.yaml
shivamerla: yes, whenever NIMs are ready for arm64

deployments/container/Dockerfile
```diff
@@ -23,10 +23,23 @@ RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o ma
 
 # Use distroless as minimal base image to package the manager binary
 FROM nvcr.io/nvidia/distroless/go:v3.1.2
+
+ARG VERSION="unknown"
```
visheshtanksale, line 27: Should we be passing the release version?

visheshtanksale (approved):
LGTM

deployments/container/Dockerfile
shivamerla: yes, it is passed from the Makefile

PR #202, opened 2024-10-31
Updating NGC home path

visheshtanksale:


slu2011 (approved):

mkhaas (commented):
lgtm. thanks @visheshtanksale

internal/controller/nimcache_controller.go
```diff
@@ -1088,7 +1088,7 @@ func (r *NIMCacheReconciler) constructJob(ctx context.Context, nimCache *appsv1a
 					},
 					{
 						Name:  "NGC_HOME",
-						Value: "/model-store/ngc",
+						Value: "/model-store/ngc/hub",
```
shivamerla, line 1091: i wonder if HF_HOME also needs to be changed for downloading vllm profile. Please double check. Also, verify that the cache is getting re-used with service deployments with this change.

shivamerla (approved):
LGTM

PR #218, opened 2024-11-06
Fix GitLab Pipeline sign

ArangoGutierrez:


.nvidia-ci.yml
```diff
@@ -180,7 +180,7 @@ sign:ngc:
 
 helm:ngc:
   image: ubuntu:latest
-  stage: release
```
shivamerla, line 183: why is the stage changed from release to sign?, helm:ngc belongs to release stage.

shivamerla: i don't get it, even with this change, i see pipeline showing same error. the order of stages is scan -> release-images -> sign-images -> helm-publish. Signing stage depends on release as expected, and helm should be on its own release-helm stage.

shivamerla (approved):

PR #239, opened 2024-11-21
Adding default node selector for NIMCache pod and job

visheshtanksale:


api/apps/v1alpha1/nimcache_types.go
```diff
@@ -39,8 +39,8 @@ type NIMCacheSpec struct {
 	Resources Resources `json:"resources,omitempty"`
 	// Tolerations for running the job to cache the NIM model
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
-	// NodeSelectors are the node selector labels to schedule the caching job.
-	NodeSelectors map[string]string `json:"gpuSelectors,omitempty"`
+	// NodeSelector is the node selector labels to schedule the caching job.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
shivamerla, line 43: hmm, this will be a breaking change. But `gpuSelectors` is confusing, we need to call this out in the documentation.

shivamerla (approved):

api/apps/v1alpha1/nimcache_types.go
visheshtanksale: Yes. Cant think any better way to do it :)

PR #262, opened 2024-12-18
Handle updated model manifest path with latest NIMs

shivamerla:
Model manifest file path has changed with recent NIMs to /opt/nim/etc/default/model_manifest.yaml. This change will handle that.

### DIFF
--- a/internal/controller/nimcache_controller.go
+++ b/internal/controller/nimcache_controller.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"reflect"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/controller/nimcache_controller.go
```diff
@@ -910,6 +911,20 @@ func getManifestConfigName(nimCache *appsv1alpha1.NIMCache) string {
 	return fmt.Sprintf("%s-manifest", nimCache.GetName())
 }
 
+func getCommand() []string {
+	// Default model manifest file path
+	defaultManifestFile := "/etc/nim/config/model_manifest.yaml"
+	updatedManifestFile := "/opt/nim/etc/default/model_manifest.yaml"
+
+	// Check for the manifest file at the updated path with latest NIMs
+	if _, err := os.Stat(updatedManifestFile); err == nil {
```
slu2011, line 920: Does the controller have access to the nim file path here?

shivamerla: Thanks Shi, i have updated this.

slu2011 (approved):

slu2011:
LGTM

PR #278, opened 2025-01-08
Adding Nemo Eval service

visheshtanksale:


api/apps/v1alpha1/nemo_evaluator_types.go
```diff
@@ -0,0 +1,764 @@
...
+		{
+			Name:  "SERVICE_ACCOUNT",
+			Value: n.Spec.ArgoWorkFlows.ServiceAccount,
+		},
+		{
+			Name:  "DATA_STORE_HOST",
+			Value: n.Spec.DataStore.Endpoint,
+		},
+		{
+			Name: "INFERENCE_URL",
```
shivamerla, line 199: does this need to be the NIM endpoint that we deployed? we also need to define a param to get this input.

api/apps/v1alpha1/nemo_evaluator_types.go
```diff
@@ -0,0 +1,764 @@
...
+// +kubebuilder:object:root=true
+
+// NemoEvaluatorList contains a list of NemoEvaluator
+type NemoEvaluatorList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []NemoEvaluator `json:"items"`
+}
+
+type Mongodb struct {
```
shivamerla, line 114: All these deps can be pulled into common types file if the params are same between all microservices.

api/apps/v1alpha1/nemo_evaluator_types.go
```diff
@@ -0,0 +1,764 @@
...
+	Scale          Autoscaling                  `json:"scale,omitempty"`
+	Metrics        Metrics                      `json:"metrics,omitempty"`
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default:=1
+	Replicas     int    `json:"replicas,omitempty"`
+	UserID       *int64 `json:"userID,omitempty"`
+	GroupID      *int64 `json:"groupID,omitempty"`
+	RuntimeClass string `json:"runtimeClass,omitempty"`
+
+	Mongodb       Mongodb       `json:"mongodb"`
```
shivamerla, line 77: These required params are going to cause a lot of pain when microservices switch to other types of DBs etc. For e.g. say version 1 of the service uses MongoDB and version 2 switches to PostgreSQL, in the version 2 CR MongoDB becomes an optional field, but the CRD says its required (otherwise existing version 1 CRs will fail).

Best way to handle this is to make all these optional in the API and enforce in the controller logic on required dependencies per microservice version and propagate as a CR status/conditions. Even better if we can add a validation webhook to verify these.

cmd/main.go
```diff
@@ -199,6 +199,17 @@ func main() {
...
 	}
 
+	if err = controller.NewNemoEvaluatorReconciler(
+		mgr.GetClient(),
+		mgr.GetScheme(),
+		updater,
+		render.NewRenderer("/manifests"),
+		ctrl.Log.WithName("controllers").WithName("NemoEvaluator"),
+	).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "NemoDatastore")
```
shivamerla, line 209: typo `NeMoDatastore`

shivamerla (commented):
Looks Great! Lets address the API changes and we can get this merged.

api/apps/v1alpha1/nemo_evaluator_types.go
visheshtanksale: Makes sense. Fixed it.

cmd/main.go
visheshtanksale: Fixed

api/apps/v1alpha1/nemo_evaluator_types.go
visheshtanksale: Fixed it!

shivamerla (approved):
LGTM!

shivamerla (approved):

PR #284, opened 2025-01-15
Draft: Nemo EntityStore API

varunrsekar:
New API for Nemo EntityStore.

Note: 
Endpoints for the different NEMO microservices need to be defined as a env variable in `spec.env`. 

Sample:
```
---
apiVersion: v1
stringData:
  password: esuser
kind: Secret
metadata:
  name: db-creds
  namespace: nemo-entity-store
type: Opaque

---
kind: NemoEntitystore
metadata:
  name: nemo-entitystore
  namespace: nemo-entity-store
spec:
  image:
    repository: registry/nemo-entity-store
    tag: latest
    pullPolicy: IfNotPresent
    pullSecrets:
      - registry-pull-secret
  authSecret: registry-api-secret
  env:
  - name: BASE_URL_DATASTORE
    value: http://datastore.nemo-datastore.svc.cluster.local
  expose:
    service:
      port: 8000
  entityStoreParams:
    appVersion: latest
    databaseName: entitystore
    databaseUser: esuser
    databaseHost: postgres.nemo-entity-store.svc.cluster.local
    databasePort: "5432"
    dbSecret: db-creds
```


api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+
+	// add standard env required for NIM service
+	envVars := []corev1.EnvVar{
+		{
+			Name:  "APP_VERSION",
+			Value: n.Spec.EntityStoreParams.AppVersion,
+		},
+		{
+			Name: "POSTGRES_PASSWORD",
```
shivamerla, line 170: isn't the key called as `POSTGRES_DB_PASSWORD` from the documentation?

api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -0,0 +1,758 @@
...
+		"app.kubernetes.io/operator-version": os.Getenv("OPERATOR_VERSION"),
+		"app.kubernetes.io/part-of":          "nemo-entitystore-service",
+		"app.kubernetes.io/managed-by":       "k8s-nim-operator",
+	}
+}
+
+// GetStandardEnv returns the standard set of env variables for the NemoEntitystore container
+func (n *NemoEntitystore) GetStandardEnv() []corev1.EnvVar {
+	passwordKey := "password"
+	if n.Spec.EntityStoreParams.DBSecret == "" {
```
shivamerla, line 159: In case of Entity Store helm chart, it creates a separate secret with the plain password provided. For operator use case, we can make an existing secret as a pre-requisite. So secret name always have to be provided. If passwordKey is not provided, then we assume the passwordKey as "password".

shivamerla, line 159: This check should be `if n.Spec.EntityStoreParams.DBSecret != ""  && n.Spec.EntityStoreParams.DBSecretPasswordKey != ""`

api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -0,0 +1,758 @@
...
+			Name:  "APP_VERSION",
+			Value: n.Spec.EntityStoreParams.AppVersion,
+		},
+		{
+			Name: "POSTGRES_PASSWORD",
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					Key: passwordKey,
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: n.Spec.EntityStoreParams.DBSecret,
```
shivamerla, line 175: We should probably throw an error in the reconciler if this field is not set.

api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -0,0 +1,758 @@
...
+	return envVars
+}
+
+// GetEnvFrom returns additional env variables to apply to the NemoEntitystore instance
+func (n *NemoEntitystore) GetEnvFrom() []corev1.EnvFromSource {
+	return []corev1.EnvFromSource{
+		{
+			ConfigMapRef: &corev1.ConfigMapEnvSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: n.Spec.EntityStoreParams.EnvConfigMap,
```
shivamerla, line 207: I am not sure if ConfigMap is required here. Since datastore is a required component, we can add a field in the API itself, (datastoreUrl) and add env with key `BASE_URL_DATASTORE`. So users can directly edit the service endpoint in the CR than maintaining a separate ConfigMap. Also, please note that, any changes to the external ConfigMap are not automatically reconciled. We would have to watch for changes on that as well.

internal/controller/nemo_entitystore_controller.go
```diff
@@ -0,0 +1,613 @@
...
+				// For other types we watch, reconcile them
+				return true
+			},
+		}).
+		Complete(r)
+}
+
+func (r *NemoEntitystoreReconciler) refreshMetrics(ctx context.Context) {
+	logger := log.FromContext(ctx)
+	// List all nodes
```
shivamerla, line 258: typo, List all instances.

api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -0,0 +1,758 @@
...
+	}
+	params.HPASpec = hpaSpec
+	return params
+}
+
+// GetSCCParams return params to render SCC from templates
+func (n *NemoEntitystore) GetSCCParams() *rendertypes.SCCParams {
+	params := &rendertypes.SCCParams{}
+	// Set metadata
+	params.Name = "nemo-guardrails-scc"
```
shivamerla, line 692: typo `nemo-entity-store-scc`

shivamerla:
thanks @varunrsekar left some comments, also some files seems to be deleted under bundle/manifests by error, please add them back.

api/apps/v1alpha1/nemo_entitystore_types.go
varunrsekar: I wanted to start a separate thread on this. We need a a proper validation-webhook to validate these CRs instead of allowing them to be created empty

shivamerla: yes, validation webhooks are ideal but mostly used for complex validations (like app version dependent checks and compatibility etc). They also provide better user experience as they fail fast on invalid params. However k8s requires webhooks to use https, hence we need to either depend on cert-manager to be deployed or manage self-signed certs manually. In the meantime, we should add validations in the controller and set proper status/conditions to indicate these error.

api/apps/v1alpha1/nemo_entitystore_types.go
shivamerla: Got it, its just that external configmap is not very intuitive and users have to manage externally. If they are part of CR, they would know what services are backing the entity-store. We can mirror what they are doing in the helm charts. If we could pass a config directly to the service as a volume mount, then an external configmap makes sense. We could compute the configmap hash and add an annotation to the deployment to rollout new revisions on changes.

```
image:
   registry: nvcr.io
   repository: nvidia/nemo-microservices/entity-store
   tag: 25.01
appConfig:
   BASE_URL_DATASTORE: http://<DATASTORE-SERVICE-NAME>.<DATASTORE-NAMESPACE>.svc.cluster.local:3000 # Update with internal DNS of Datastore microservice
```

api/apps/v1alpha1/nemo_entitystore_types.go
varunrsekar: I verified that the key is `POSTGRES_PASSWORD`

api/apps/v1alpha1/nemo_entitystore_types.go
varunrsekar: yeah makes sense. The `DBSecret` would be a required field and if `DBSecretPasswordKey` is provided, we would use that, toherwise the default key is `password`

varunrsekar: Just `n.Spec.EntityStoreParams.DBSecretPasswordKey != ""` should suffice here. I'll add the validation for `DBSecret` in the controller

api/apps/v1alpha1/nemo_entitystore_types.go
varunrsekar: I just noticed there's a `spec.env` field. Should I use that instead to make these env variables and skip the field inside the params?

shivamerla: Just to clarify, in the CR spec, we can take this as a opaque config (map[string]string) and generate env key-pairs out of it. So we don't have to change the API for every new service.

```
AppConfig map[string]string `json:"appConfig"`
```

shivamerla: Yes, that works too, we could just make them as env variables and add them in CR samples and docs.

api/apps/v1alpha1/nemo_entitystore_types.go
varunrsekar: I've reworked the validations in a new commit.

shivamerla (approved):
LGTM!

PR #286, opened 2025-01-16
Support DataStore V2

shengnuo:
New API for supporting DataStore V2

Sample CR can be found at `config/samples/apps_v1alpha1_nimdatastorev2.yaml`

pre-requisite components:
- Minio
- Postgres

Secrets as specified in `.spec.dataStoreParams` that need to be created before deploying DataStore microservice
- datastoreConfigSecret
- datastoreInitSecret
- datastoreInlineConfigSecret
- datastoreSettingsSecret
- dbSecret
- giteaAdminSecret
- lfsJwtSecret
- objStoreSecret


slu2011 (approved):
LGTM

api/apps/v1alpha1/nemo_datastorev2_types.go
```diff
@@ -0,0 +1,1046 @@
...
+}
+
+// NemoDatastoreV2Status defines the observed state of NemoDatastoreV2
+type NemoDatastoreV2Status struct {
+	Conditions        []metav1.Condition `json:"conditions,omitempty"`
+	AvailableReplicas int32              `json:"availableReplicas,omitempty"`
+	State             string             `json:"state,omitempty"`
+}
+
+type NemoDatastoreV2Params struct {
```
shivamerla, line 91: we should split these into "db", "objectStore", "storage", "sshService" types. And map underlying values as done in the helm chart. Params itself need not be same as helm chart values.

api/apps/v1alpha1/nemo_datastorev2_types.go
```diff
@@ -0,0 +1,1046 @@
...
+			Name:      "inline-config-sources",
+		},
+		{
+			MountPath: "/usr/sbin/init",
+			Name:      "init",
+		},
+	}
+}
+
+func (n *NemoDatastoreV2) GetInitContainers() []corev1.Container {
```
shivamerla, line 647: currently we are not rendering initContainers in the deployment/statefulset manifests, we need to add that logic.

internal/controller/nemo_datastorev2_controller.go
```diff
@@ -0,0 +1,642 @@
...
+	logger := log.FromContext(ctx)
+	var err error
+	defer func() {
+		if err != nil {
+			r.GetEventRecorder().Eventf(nemoDatastore, corev1.EventTypeWarning, conditions.Failed,
+				"NemoDatastoreV2 %s failed, msg: %s", nemoDatastore.Name, err.Error())
+		}
+	}()
+
+	pvc := corev1.PersistentVolumeClaim{}
```
shivamerla, line 280: pull this into a separate method `getPVCSpec()`

shivamerla: we should consider following fields to be updated in the spec from the `storage` params

```
persistence:
  enabled: true
  create: true
  existingClaim: datastore-shared-storage
  size: 5Gi
  accessModes:
    - ReadWriteOnce
  labels: {}
  annotations: {}
  storageClass:
  subPath:
```

internal/controller/nemo_datastorev2_controller.go
```diff
@@ -0,0 +1,642 @@
...
+	}
+	pvc.Spec.Resources = corev1.VolumeResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceStorage: size,
+		},
+	}
+	pvc.Spec.AccessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOnce}
+	pvc.Namespace = nemoDatastore.Namespace
+
+	err = r.Create(ctx, &pvc)
```
shivamerla, line 295: This needs to be called only when `storage.pvc.create` is set.

api/apps/v1alpha1/nemo_datastorev2_types.go
```diff
@@ -0,0 +1,1046 @@
...
+	// NemoDatastoreV2StatusNotReady indicates that NEMO datastore service is not ready
+	NemoDatastoreV2StatusNotReady = "NotReady"
+	// NemoDatastoreV2StatusReady indicates that NEMO datastore service is ready
+	NemoDatastoreV2StatusReady = "Ready"
+	// NemoDatastoreV2StatusFailed indicates that NEMO datastore service has failed
+	NemoDatastoreV2StatusFailed = "Failed"
+)
+
+// NemoDatastoreV2Spec defines the desired state of NemoDatastoreV2
+type NemoDatastoreV2Spec struct {
```
shivamerla, line 55: as discussed please change this into just `NemoDatastore` type.

api/apps/v1alpha1/nemo_datastorev2_types.go
shengnuo: See `internal/controller/nemo_datastorev2_controller.go:L382`

api/apps/v1alpha1/nemo_datastorev2_types.go
shengnuo: It's difficult to categorize the pre-existing secrets into these categorize like the NeMo datastore helmchart..
Some of the pre-existing secrets come with the helmcharts, and they don't seem to be tied to neither "db", "objectStore", "storage", nor "sshService. For example, `datastoreInitSecret` contains the shell script for initializing datastore volumes.

We can discuss further during the demo

shengnuo: Another option would be to make the NeMo datastore controller fetch and deploy the nemo datastore helmchart... so that the datastore CR can be formatted exactly like the helm chart values file

shivamerla: no, we will be tied to the helm spec in that case. We can further discuss on those params and see if we can avoid them while deployment.

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -88,20 +89,31 @@ type NemoDatastoreStatus struct {
...
+	DataStoreInitSecret         string `json:"datastoreInitSecret"`
+	DataStoreConfigSecret       string `json:"datastoreConfigSecret"`
+	DataStoreInlineConfigSecret string `json:"datastoreInlineConfigSecret"`
+
+	SshEnabled bool `json:"sshEnabled"`
+
+	Persistence *Persistence `json:"persistence,omitempty"`
+}
+
+type Persistence struct {
```
shivamerla, line 109: lets use the pvc spec we have defined [here](https://github.com/NVIDIA/k8s-nim-operator/blob/main/api/apps/v1alpha1/nimcache_types.go#L132) to be consistent

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -88,20 +89,31 @@ type NemoDatastoreStatus struct {
...
 
-	InitContainerImage   string   `json:"initContainerImage,omitempty"`
-	InitContainerCommand []string `json:"initContainerCommand,omitempty"`
+	DataStoreInitSecret         string `json:"datastoreInitSecret"`
+	DataStoreConfigSecret       string `json:"datastoreConfigSecret"`
+	DataStoreInlineConfigSecret string `json:"datastoreInlineConfigSecret"`
+
+	SshEnabled bool `json:"sshEnabled"`
+
+	Persistence *Persistence `json:"persistence,omitempty"`
```
shivamerla, line 106: we should rename this into just `PVC` and json tag as `pvc` as it is well known.

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -157,97 +169,319 @@ func (n *NemoDatastore) GetStandardLabels() map[string]string {
...
-	}
-	cmd := n.Spec.DataStoreParams.InitContainerCommand
-	if len(cmd) == 0 {
-		cmd = []string{
-			"sh",
-			"-c",
-			fmt.Sprintf("until nc -z %s %s; do echo \"PostgreSQL is unavailable. Sleeping for 5 seconds\"; sleep 5; done;", n.Spec.DataStoreParams.DatabaseHost, n.Spec.DataStoreParams.DatabasePort),
-		}
+// GetPVC returns the persistence PVC object for the NeMoDatastore container
+func (n *NemoDatastore) GetPVC() (*corev1.PersistentVolumeClaim, error) {
```
shivamerla, line 433: we can instead use shared.ConstructPVC() method directly in the controller package: https://github.com/NVIDIA/k8s-nim-operator/blob/main/internal/shared/pvc.go#L29

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -157,97 +169,319 @@ func (n *NemoDatastore) GetStandardLabels() map[string]string {
...
+			},
+		},
+	}, nil
+}
+
+func (n *NemoDatastore) ShouldCreatePersistentStorage() bool {
+	return n.Spec.DataStoreParams.Persistence != nil && n.Spec.DataStoreParams.Persistence.Create
+}
+
+func (n *NemoDatastore) ShouldEnablePersistentStorage() bool {
```
shivamerla, line 467: i think we don't need `Enabled` field at all, if pvc is nil in the spec, we can assume it is not enabled. If provided in the spec, either an pvc name should to be provided or create option have to be enabled. similar checks are done [here](https://github.com/NVIDIA/k8s-nim-operator/blob/main/internal/controller/nimcache_controller.go#L491).

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -429,15 +663,88 @@ func (n *NemoDatastore) GetDefaultStartupProbe() *corev1.Probe {
...
+		},
+		{
+			MountPath: "/usr/sbin/init",
+			Name:      "init",
+		},
+	}
+}
+
+func (n *NemoDatastore) GetInitContainers() []corev1.Container {
+	user := int64(1000)
```
shivamerla, line 706: here, we can use the same RunAsUser config from the main container if provided. If not set then default to 1000. The reason in in some distributions, the namespace is limited to certain range of UIDs and this might fail then.

bundle/manifests/apps.nvidia.com_nemodatastores.yaml
```diff
@@ -65,40 +65,60 @@ spec:
...
+                        type: object
+                      size:
+                        type: string
+                      storageClass:
+                        type: string
+                    required:
+                    - claimName
+                    - create
+                    - enabled
+                    - size
```
shivamerla, line 108: size is not a required param if this is an existing PVC. Same with claimName, if creation is enabled, we can add with default claimname based on the CR instance name.

internal/controller/nemo_datastore_controller.go
```diff
@@ -375,6 +387,10 @@ func (r *NemoDatastoreReconciler) reconcileNemoDatastore(ctx context.Context, ne
 		if len(envFrom) > 0 {
 			result.Spec.Template.Spec.Containers[0].EnvFrom = envFrom
 		}
+		fsGroup := int64(1000)
```
shivamerla, line 390: same here, if RunAsUser is set then we should use that instead.

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -78,7 +80,19 @@ type NemoDatastoreSpec struct {
 	GroupID      *int64 `json:"groupID,omitempty"`
 	RuntimeClass string `json:"runtimeClass,omitempty"`
 
-	DataStoreParams NemoDataStoreParams `json:"dataStoreParams"`
+	ObjectStore      ObjectStore      `json:"objectStore"`      // e.g. minio
+	ExternalDatabase ExternalDatabase `json:"externalDatabase"` // e.g. postgres
+
+	Secrets Secrets                `json:"secrets"`
+	PVC     *PersistentVolumeClaim `json:"pvc,omitempty"`
```
varunrsekar, line 87: Perhaps the definition of this struct needs to be moved to `common_types.go`. Its currently residing in `nimcache_types.go`

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -88,20 +102,26 @@ type NemoDatastoreStatus struct {
...
-	InitContainerCommand []string `json:"initContainerCommand,omitempty"`
+type ExternalDatabase struct {
+	SSLMode           string `json:"sslMode"`
+	Host              string `json:"host"`
+	Port              int    `json:"port"`
+	User              string `json:"user"`
+	Database          string `json:"database"`
+	DatabaseSecret    string `json:"databaseSecret"`
+	DatabaseSecretKey string `json:"databaseSecretKey"`
 }
```
varunrsekar, line 125: Can you follow this PR and use DatabaseConfig for this object? https://github.com/NVIDIA/k8s-nim-operator/pull/300/files

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -88,20 +102,26 @@ type NemoDatastoreStatus struct {
...
-	GiteaEndpoint string `json:"giteaEndpoint"`
-	GiteaSecret   string `json:"giteaSecret"`
-	DatabaseURL   string `json:"databaseURL"`
-	DatabaseHost  string `json:"databaseHost"`
-	DatabasePort  string `json:"databasePort"`
-	DBSecret      string `json:"dbSecret"`
+type ObjectStore struct { // e.g. Minio, s3
+	ObjectStoreSecret             string `json:"objectStoreSecret"`
+	ObjectStoreAccessKey          string `json:"objectStoreSecretAccessKey"`
+	ObjectStoreSecretAccessSecret string `json:"objectStoreSecretAccessSecret"`
```
varunrsekar, line 108: This could be defined in a separate object for `credentials` (ObjectStoreCredentials)

api/apps/v1alpha1/nemo_datastore_types.go
shengnuo: This would affect the nimcache service as well
Best to handle that in a separate PR

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -88,20 +105,49 @@ type NemoDatastoreStatus struct {
...
 
-	EnvConfigMap string `json:"envConfigmap"`
-	EnvSecret    string `json:"envSecret"`
+	// PasswordKey is the name of the key in the `CredentialsSecret` secret for the object store credentials.
+	PasswordKey string `json:"passwordKey"`
+}
 
-	InitContainerImage   string   `json:"initContainerImage,omitempty"`
-	InitContainerCommand []string `json:"initContainerCommand,omitempty"`
+type ExternalDatabase struct {
```
visheshtanksale, line 136: Where is this object used?

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -157,99 +203,306 @@ func (n *NemoDatastore) GetStandardLabels() map[string]string {
...
-		cmd = []string{
-			"sh",
-			"-c",
-			fmt.Sprintf("until nc -z %s %s; do echo \"PostgreSQL is unavailable. Sleeping for 5 seconds\"; sleep 5; done;", n.Spec.DataStoreParams.DatabaseHost, n.Spec.DataStoreParams.DatabasePort),
-		}
-	}
-	return []corev1.Container{
+// GetVolumes returns volumes for the NemoDatastore container
+func (n *NemoDatastore) GetVolumes() []corev1.Volume {
+	/*volumes:
```
visheshtanksale, line 421: Is this comment needed?

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -157,99 +203,306 @@ func (n *NemoDatastore) GetStandardLabels() map[string]string {
...
+		})
+	}
+	return volumes
+}
+
+func (n *NemoDatastore) ShouldCreatePersistentStorage() bool {
+	return n.Spec.PVC != nil && n.Spec.PVC.Create != nil && *n.Spec.PVC.Create
+}
+
+// GetStandardAnnotations returns default annotations to apply to the NemoDatastore instance
```
visheshtanksale, line 503: Fix the method comment.

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -88,20 +105,49 @@ type NemoDatastoreStatus struct {
...
 
-type NemoDataStoreParams struct {
-	AppVersion    string `json:"appVersion"`
-	GiteaEndpoint string `json:"giteaEndpoint"`
-	GiteaSecret   string `json:"giteaSecret"`
-	DatabaseURL   string `json:"databaseURL"`
-	DatabaseHost  string `json:"databaseHost"`
-	DatabasePort  string `json:"databasePort"`
-	DBSecret      string `json:"dbSecret"`
+type ObjectStoreConfig struct { // e.g. Minio, s3
```
visheshtanksale, line 108: Shouldn't some of these fields under `ObjectStoreConfig`  be required?

visheshtanksale (commented):
Some minor comments.

api/apps/v1alpha1/nemo_datastore_types.go
shengnuo: It's now replaced by `DatabaseConfig` in nemo_common_types.go`

Removing the struct as it's no longer used

api/apps/v1alpha1/nemo_datastore_types.go
shengnuo: Remove the function as it's no longer used

api/apps/v1alpha1/nemo_datastore_types.go
shengnuo: Ya they are required if there's the json tag does not have omitempty
See config/crd/bases/apps.nvidia.com_nemodatastores.yaml:L809-813

```
  required:
  - bucketName
  - credentials
  - endpoint
  - region
  - ssl
 ```

api/apps/v1alpha1/nemo_datastore_types.go
shengnuo: It helps folks to see what the generated volume would look like in the deployment manifest IMO

visheshtanksale (approved):
LGTM

shivamerla (approved):
looks good

PR #298, opened 2025-01-22
refactor entitystore CRD

varunrsekar:
This PR reorganizes the EntityStore CR for better readability. Also removed the code validations in favor of input validations in CRD.


shivamerla (approved):
LGTM

PR #303, opened 2025-01-28
fix stale nemo datastore CRDs

varunrsekar:
This change adds the missing diffs from running `make sync-crds`


varunrsekar:
> LGTM! Commit needs signature

The signature was there. But had to fix my git config. The error is gone now

shivamerla (approved):

PR #308, opened 2025-01-31
Fix project scaffolding for all microservices

shivamerla:


varunrsekar:
QQ: Do we need a separate section for `update` verb on `<obj>/finalizers` resources? The `k8s-nim-operator-role` RBAC in the helm chart mentions that explicitly.

shivamerla:
> QQ: Do we need a separate section for `update` verb on `<obj>/finalizers` resources? The `k8s-nim-operator-role` RBAC in the helm chart mentions that explicitly.

Yes, explicit permissions are required to update finalizers, as they are a subresource similar to status. Auto-generation does not include these rules by default, leading to errors without them.

slu2011 (approved):

PROJECT
```diff
@@ -41,7 +41,38 @@ resources:
   controller: true
   domain: nvidia.com
   group: apps
+<<<<<<< HEAD
```
mikemckiernan, line 44: unsolicited review: I see several of these rebase-y-looking lines in a few files, FYI.

shivamerla: Good reviews always welcome! Thanks for pointing this out, i have fixed the rebase errors.

varunrsekar:
> > QQ: Do we need a separate section for `update` verb on `<obj>/finalizers` resources? The `k8s-nim-operator-role` RBAC in the helm chart mentions that explicitly.
> 
> Yes, explicit permissions are required to update finalizers, as they are a subresource similar to status. Auto-generation does not include these rules by default, leading to errors without them.

I was confused because I don't see the RBAC for finalizers in this PR. Is that expected?

shivamerla:
> > > QQ: Do we need a separate section for `update` verb on `<obj>/finalizers` resources? The `k8s-nim-operator-role` RBAC in the helm chart mentions that explicitly.
> > 
> > 
> > Yes, explicit permissions are required to update finalizers, as they are a subresource similar to status. Auto-generation does not include these rules by default, leading to errors without them.
> 
> I was confused because I don't see the RBAC for finalizers in this PR. Is that expected?

These are already auto-generated and in the config/rbac/role.yaml [here](https://github.com/NVIDIA/k8s-nim-operator/blob/main/config/rbac/role.yaml#L81). The editor/viewer roles doesn't include them.

mkhaas (commented):
LGTM

varunrsekar:
LGTM

visheshtanksale (approved):
LGTM

PR #316, opened 2025-02-07
Add support to configure multiple service ports

shivamerla:
* Default named port of "api" will be added with legacy .spec.port attribute
 * Ability to setup multiple named ports "api", "metrics" etc.
 * auto detection of metrics and api ports for probes and metrics collection


api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -510,6 +506,25 @@ func (n *NemoEntitystore) GetDeploymentParams() *rendertypes.DeploymentParams {
 
 	// Set runtime class
 	params.RuntimeClassName = n.GetRuntimeClass()
+
+	// Extract ports from spec and update rendering params
+	if len(n.GetServicePorts()) > 0 {
+		var containerPorts []corev1.ContainerPort
+		for _, svcPort := range n.GetServicePorts() {
```
varunrsekar, line 513: common nit across files:
```suggestion
        svcPorts := n.GetServicePorts()
	if len(svcPorts) > 0 {
		var containerPorts []corev1.ContainerPort
		for _, svcPort := range svcPorts {
```

api/apps/v1alpha1/common_types.go
```diff
@@ -115,3 +116,59 @@ type CertConfig struct {
...
+			if port.Name == name {
+				return name
+			}
+		}
+	}
+	return ""
+}
+
+// getProbePort determines the appropriate port for probes based on the service spec.
+func getProbePort(serviceSpec Service) intstr.IntOrString {
```
varunrsekar, line 133: we can probably use `rendertypes.ServiceParams` as input here and simplify this to always prefer "api" namedPort from the `ServiceParams.Ports` list

api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -510,6 +506,25 @@ func (n *NemoEntitystore) GetDeploymentParams() *rendertypes.DeploymentParams {
...
+				Protocol:      svcPort.Protocol,
+				ContainerPort: svcPort.Port,
+			})
+		}
+		params.Ports = containerPorts
+	} else {
+		params.Ports = []corev1.ContainerPort{{
+			Name:          "api",
+			Protocol:      corev1.ProtocolTCP,
+			ContainerPort: n.GetServicePort(),
```
varunrsekar, line 525: It feels weird to introduce a new API "Ports" but depend on a deprecated field "Port" to get the default value. We should just explicitly set `8000` here

varunrsekar (changes requested):

api/apps/v1alpha1/nemo_entitystore_types.go
shivamerla: Makes sense, i was thinking of adding a default in the code when we remove the current `Port` API from the spec. Currently that is set to default in the CRD itself, so we always have it until we remove that from the API.

varunrsekar: yeah better to do it now. Lets have new code assume deprecated apis don't exist. Otherwise removal becomes a chore.

config/samples/apps_v1alpha1_nemoguardrails.yaml
```diff
@@ -8,56 +8,53 @@ metadata:
...
-	pullSecrets:
-	- ngc-secret
-	repository: nvcr.io/nvidian/nemo-llm/nemo-guardrails-microservice
-	tag: "25.02-rc4"
+    service:
+      type: ClusterIP
+      ports:
+        - name: api
+          port: 7331
+  pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
shivamerla: Thanks for catching this @a-mccarthy The indentation was all messed up in this CR sample, will add the image field with right indentation.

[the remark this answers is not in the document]
shivamerla: fixed it.

api/apps/v1alpha1/nemo_entitystore_types.go
shivamerla: done

api/apps/v1alpha1/nemo_entitystore_types.go
shivamerla: done

api/apps/v1alpha1/common_types.go
shivamerla: Each resource is rendered independently during reconciliation. That will add a dependency that we always have to render serviceParams before setting up deployment or servicemonitor.

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -592,10 +593,7 @@ func (n *NemoDatastore) GetDefaultLivenessProbe() *corev1.Probe {
 		ProbeHandler: corev1.ProbeHandler{
 			HTTPGet: &corev1.HTTPGetAction{
 				Path: "/v1/health/live",
-				Port: intstr.IntOrString{
-					Type:   intstr.Type(0),
-					IntVal: n.Spec.Expose.Service.Port,
-				},
+				Port: getProbePort(n.Spec.Expose.Service),
```
shengnuo, line 596: This implies that `.spec.expose.service` is required, but it is not currently.
Maybe we should make it a required field or add a default value?

api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -649,10 +644,7 @@ func (n *NemoDatastore) GetDefaultStartupProbe() *corev1.Probe {
 		ProbeHandler: corev1.ProbeHandler{
 			HTTPGet: &corev1.HTTPGetAction{
 				Path: "/v1/health/ready",
-				Port: intstr.IntOrString{
-					Type:   intstr.Type(0),
-					IntVal: n.Spec.Expose.Service.Port,
-				},
+				Port: getProbePort(n.Spec.Expose.Service),
```
shengnuo, line 647: `GetDefaultStartupProbe` should be called by `GetStartupProbe`
It isn't the case currently, we should make that change

api/apps/v1alpha1/common_types.go
```diff
@@ -115,3 +116,59 @@ type CertConfig struct {
...
+		if port.Name != "" {
+			return intstr.FromString(port.Name)
+		}
+		return intstr.FromInt(int(port.Port))
+	case 0:
+		// Default to "api" as the operator always adds a default named port with 8000
+		return intstr.FromString("api")
+	default:
+		// Multiple ports: Prefer "api"
+		if portName := selectNamedPort(serviceSpec, "api"); portName != "" {
```
shengnuo, line 146: nit: consider wrapping the default `api` port name in a constant

api/apps/v1alpha1/nemo_evaluator_types.go
shengnuo, line 1: Evaluator currently has no default start up probe. Should we add a default like other services?

api/apps/v1alpha1/common_types.go
```diff
@@ -115,3 +116,59 @@ type CertConfig struct {
...
+	case 0:
+		// Default to "api" as the operator always adds a default named port with 8000
+		return intstr.FromString("api")
+	default:
+		// Multiple ports: Prefer "metrics", fallback to "api"
+		if portName := selectNamedPort(serviceSpec, "metrics", "api"); portName != "" {
+			return intstr.FromString(portName)
+		}
+		// Default when multiple ports exist
+		return intstr.FromString("metrics")
```
varunrsekar, line 172: similarly here: Shouldn't we be getting the first port from the list if "metrics" and "api" dont exist?

api/apps/v1alpha1/common_types.go
```diff
@@ -115,3 +116,59 @@ type CertConfig struct {
...
+		return intstr.FromString("api")
+	default:
+		// Multiple ports: Prefer "api"
+		if portName := selectNamedPort(serviceSpec, "api"); portName != "" {
+			return intstr.FromString(portName)
+		}
+		// Default when multiple ports exist
+		return intstr.FromString("api")
+	}
+}
```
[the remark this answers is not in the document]
shivamerla: We just can't assume to use the first port when multiple ports are given. For e.g. in case of Customizer, if "api" and "internal" ports are used with different names, we can't assume "internal" ports for liveness/startup probes etc. Hence i was always enforcing that a named port of "api" should be specified in that case. Probably better things is to validate this during reconciliation and error out when required named ports are not provided.

api/apps/v1alpha1/common_types.go
shivamerla: same logic as above, in documentation we have to specify that named ports for "api" or "metrics" should be provided when they are served differently by the underlying service. Randomly assigning would not work.

api/apps/v1alpha1/nemo_datastore_types.go
shivamerla: yes, we can make this as a required field, earlier there was just single port param and default for that, so it was always present.

api/apps/v1alpha1/common_types_test.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func TestGetMetricsPort(t *testing.T) {
+	tests := []struct {
+		name        string
+		serviceSpec Service
+		expected    intstr.IntOrString
+	}{
+		{"Single named port", Service{Ports: []corev1.ServicePort{{Name: "metrics", Port: 8081}}}, intstr.FromString("metrics")},
+		{"Single unnamed port", Service{Ports: []corev1.ServicePort{{Port: 8181}}}, intstr.FromInt(8181)},
```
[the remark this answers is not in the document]
shivamerla: when say user provides input as below:

```
  expose:
    service:
      type: ClusterIP
      ports:
        - port: 8000
           protocol: TCP
```

then, in the deployment spec, we don't add any named ports like "api". Hence we use the port number that is given to us. we will be add a default named port ("api: 8000" ) when no ports are given in the config at all.

api/apps/v1alpha1/common_types.go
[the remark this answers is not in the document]
varunrsekar: But arent we just taking whatever port there is when there's just a single port given? Whats the difference in that cae? Also, if we hard-require "api" port, then we can simply return "api"nwithout any of this logic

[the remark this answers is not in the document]
shivamerla: These are the scenarios possible:

1. No explicit port is provided in the config
In this case, we created a default named port of "api" with value "8000" and setup service,  liveness/readiness/startup probes, and metrics port in the servicemonitor.

2. Single port is provided in the config

```
  expose:
    service:
      type: ClusterIP
      ports:
        - port: 4317
           protocol: TCP
```
In this case, we setup an unnamed port "4317" in the service, liveness/readiness/startup and servicemonitor configs.

```
  expose:
    service:
      type: ClusterIP
      ports:
        - port: 8000
           protocol: TCP
           name: http
````

In this case, we setup a named port "http" in the service, liveness/readiness/startup and servicemonitor configs.

4. Multiple ports are defined for the config

```
  expose:
    service:
      type: ClusterIP
      ports:
        - port: 8000
           protocol: TCP
           name: http

  expose:
    service:
      type: ClusterIP
      ports:
        - port: 9000
           protocol: TCP
           name: metrics
```

In this case, we cannot derive ports to use for liveness/readiness/startup probes, so we enforce that "api" named port be defined. For servicemonitor config we use the "metrics" named port.

```
  expose:
    service:
      type: ClusterIP
      ports:
        - port: 4317
           protocol: TCP

  expose:
    service:
      type: ClusterIP
      ports:
        - port: 4317
           protocol: TCP
```

In this case, we cannot derive ports to use for liveness/readiness/startup probes and servicemonitor, so we enforce that "api" and "metrics" named ports be defined.

```
  expose:
    service:
      type: ClusterIP
      ports:
        - port: 8000
           protocol: TCP
           name: api

  expose:
    service:
      type: ClusterIP
      ports:
        - port: 9000
           protocol: TCP
           name: metrics
```

This is the ideal case, where we setup both probes and metrics ports correctly.

File: api/apps/v1alpha1/common_types.go [hunk 11]
```diff
@@ -34,13 +43,11 @@ type Service struct {
...
-	// +kubebuilder:deprecatedversion
-	// +kubebuilder:default=8000
-	Port int32 `json:"port,omitempty"`
-	// Defines multiple ports for the service
-	Ports       []corev1.ServicePort `json:"ports,omitempty"`
-	Annotations map[string]string    `json:"annotations,omitempty"`
+	// Port is the main api serving port
+	Port int32 `json:"port"`
+	// MetricsPort is the port to be used for the metrics collection (optional)
+	MetricsPort int32             `json:"metricsPort,omitempty"`
```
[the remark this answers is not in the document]
varunrsekar:
```suggestion
	// Port is the main api serving port
	Port *int32 `json:"port"`
	// MetricsPort is the port to be used for the metrics collection (optional)
	MetricsPort *int32             `json:"metricsPort,omitempty"`
```
Also, these ports needs to be a pointer. Not setting `port` and setting `port: 0` have different meanings.

api/apps/v1alpha1/nemo_customizer_types.go
```diff
@@ -435,18 +441,15 @@ func (n *NemoCustomizer) GetStartupProbe() *corev1.Probe {
 // GetDefaultStartupProbe returns the default startup probe for the NemoCustomizer container
 func (n *NemoCustomizer) GetDefaultStartupProbe() *corev1.Probe {
```
varunrsekar, line 442: Can we extract this to a common util? The only difference for each CR seems to be the `Path`

api/apps/v1alpha1/nemo_customizer_types.go
```diff
@@ -668,23 +642,27 @@ func (n *NemoCustomizer) GetDeploymentParams() *rendertypes.DeploymentParams {
...
+			ContainerPort: CustomizerAPIPort,
+		},
+		{
+			Name:          DefaultNamedPortInternal,
+			Protocol:      corev1.ProtocolTCP,
+			ContainerPort: CustomizerInternalPort,
+		},
+	}
+
+	if n.GetMetricsPort() != 0 {
```
varunrsekar, line 659: For nimservice, defauling `MetricsPort` to `9009` would make sense since they always expose it. But for nemo microservices, they dont yet have metrics exposed in a separate port. How do we differentiate between the 2 cases?

Do we still need to default the metrics port here?

varunrsekar (changes requested):

api/apps/v1alpha1/common_types.go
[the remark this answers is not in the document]
shivamerla: No, internal port need not be configurable, this is for the jobs to report back status to the customer service and all auto-configured by the operator. Also, the port setting was intentional. We don't want users to be able to set to `0`.  In that case we use the default port `8000`.

api/apps/v1alpha1/nemo_customizer_types.go
shivamerla: This is true for all defaults being applied, need to refactor in a separate MR.

api/apps/v1alpha1/nemo_customizer_types.go
shivamerla: NIMs don't expose 9009, all are reported through default API port. Same with NeMo services too.

api/apps/v1alpha1/common_types.go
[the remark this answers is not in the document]
varunrsekar: > We don't want users to be able to set to 0
To address this, we would add a validation for that value to be between `1` and `65535`.

k8s APIs explicitly reject that port number to be set to be `0`. We should atleast be consistent with that. But for the per-service defaults, this still needs to be a ptr.

api/apps/v1alpha1/nemo_customizer_types.go
varunrsekar: Was `metrics` port introduced as a placeholder? Trying to understand which CR uses a separate port for metrics. If no one is using it, lets remove it now and reintroduce it later?

api/apps/v1alpha1/common_types.go
```diff
@@ -23,6 +23,15 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 )
 
+const (
+	// DefaultAPIPort is the default api  port
+	DefaultAPIPort = 8000
+	// DefaultNamedPortAPI is the default name for api port
+	DefaultNamedPortAPI = "api"
+	// DefaultNamedPortMetrics is the default name for metrics port
+	DefaultNamedPortMetrics = "metrics"
```
varunrsekar, line 32: nit: cleanup

api/apps/v1alpha1/nemo_entitystore_types.go
```diff
@@ -38,6 +38,8 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 const (
+	// EntitystoreAPIPort is the default port that the entittystore serves on
```
varunrsekar, line 41:
```suggestion
	// EntitystoreAPIPort is the default port that the entitystore serves on
```

api/apps/v1alpha1/nemo_evaluator_types.go
```diff
@@ -684,9 +712,15 @@ func (n *NemoEvaluator) GetServiceParams() *rendertypes.ServiceParams {
 	params.Type = "ClusterIP"
```
varunrsekar, line 712: probably not for this PR but a nit:
```suggestion
	params.Type = n.GetServiceType()
```

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -768,22 +779,16 @@ func (n *NIMService) GetServiceMonitorParams() *rendertypes.ServiceMonitorParams
...
-		},
-		Selector: metav1.LabelSelector{
-			MatchLabels: n.GetServiceLabels(),
-		},
+		NamespaceSelector: monitoringv1.NamespaceSelector{MatchNames: []string{n.Namespace}},
+		Selector:          metav1.LabelSelector{MatchLabels: n.GetServiceLabels()},
 		Endpoints: []monitoringv1.Endpoint{
 			{
-				Port:          "service-port",
-				Path:          "/v1/metrics",
```
varunrsekar, line 780: Was this supposed to be removed? The default value for `path` is `/metrics` while nimservice exposes `/v1/metrics`

manifests/service.yaml
```diff
@@ -3,37 +3,36 @@ kind: Service
...
+    {{- else if .Port }}
     - port: {{ .Port }}
       {{- if .TargetPort }}
       targetPort: {{ .TargetPort }}
       {{- end }}
-      {{- if .Protocol }}
-      protocol: {{ .Protocol }}
-      {{- end }}
-      {{- if .PortName }}
-      name: {{ .PortName }}
```
varunrsekar, line 37: Can you also clean up `ServiceParams` in `internal/render/types`? Looks like `PortName` is no longer being used from there

varunrsekar (changes requested):

api/apps/v1alpha1/nimservice_types.go
shivamerla: good catch! this was not supposed to be removed

varunrsekar (approved):
LGTM

PR #318, opened 2025-02-11
Add NemoServices API for single click management

shivamerla:


internal/controller/nemoservice_controller.go
```diff
@@ -0,0 +1,586 @@
...
+
+	// Set service spec
+	err := SetObjectSpec(obj, service)
+	if err != nil {
+		logger.Error(err, "Failed to set object spec")
+		return err
+	}
+
+	// Set controller reference
+	if err := controllerutil.SetControllerReference(nemoService, obj, r.Scheme); err != nil {
```
varunrsekar, line 304: I believe we should wait for all child CRs to be deleted before the parent CR is deleted here.
```suggestion
	if err := controllerutil.SetControllerReference(nemoService, obj, r.Scheme, controllerutil.WithBlockOwnerDeletion(true)); err != nil {
```

internal/controller/nemoservice_controller.go
```diff
@@ -0,0 +1,586 @@
...
+		return result, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *NemoServiceReconciler) cleanupNemoService(ctx context.Context, nemoService *appsv1alpha1.NemoService) error {
+	// TODO: Add custom cleanup
+
+	// All owned objects are garbage collected
```
varunrsekar, line 139: Shouldn't we wait for all child CRs to be cleaned up before this CR is deleted? Only Kubernetes-managed objects automatically block their parent CRs from being deleted before them

internal/controller/nemoservice_controller.go
```diff
@@ -0,0 +1,586 @@
...
+
+	// Define service reconcilers
+	serviceReconcilers := map[ServiceType]func(context.Context, *NemoServiceReconciler, *appsv1alpha1.NemoService) error{
+		Customizer: func(ctx context.Context, r *NemoServiceReconciler, ns *appsv1alpha1.NemoService) error {
+			spec := NemoServiceSpec[appsv1alpha1.NemoServiceCustomizerSpec]{
+				Enabled: ns.Spec.Customizer.Enabled,
+				Name:    ns.Spec.Customizer.Name,
+				Spec:    ns.Spec.Customizer,
+			}
+			enabledServices[spec.Name] = *spec.Enabled
```
varunrsekar, line 158: nit here and other references below:
Need to check for nil before dereferencing this ptr.
```suggestion
			enabledServices[spec.Name] = spec.Enabled != nil && *spec.Enabled
```

internal/controller/nemoservice_controller.go
```diff
@@ -0,0 +1,586 @@
...
+			return reconcileGenericService(ctx, r, ns, spec, &appsv1alpha1.NemoEntitystore{})
+		},
+	}
+
+	// Reconcile all enabled services
+	for serviceType, reconcileFunc := range serviceReconcilers {
+		serviceSpec := getServiceSpec(nemoService, serviceType)
+		if serviceSpec.Enabled != nil && *serviceSpec.Enabled {
```
varunrsekar, line 202: We can simplify with `enabledServices` here if we key-in using `serviceType` in that map. That will eliminate the need for `getServiceSpec`

internal/controller/nemoservice_controller.go
```diff
@@ -0,0 +1,586 @@
...
+	service NemoServiceSpec[T],
+	obj client.Object,
+) error {
+	logger := log.FromContext(ctx)
+
+	if service.Enabled == nil || !*service.Enabled {
+		return nil
+	}
+
+	// Set metadata
```
varunrsekar, line 292: We should also set a label/annotation that links it to the parent CR.

internal/controller/nemoservice_controller.go
```diff
@@ -0,0 +1,586 @@
...
+		Customizer:  &appsv1alpha1.NemoCustomizerList{},
+		Datastore:   &appsv1alpha1.NemoDatastoreList{},
+		Evaluator:   &appsv1alpha1.NemoEvaluatorList{},
+		Guardrails:  &appsv1alpha1.NemoGuardrailList{},
+		Entitystore: &appsv1alpha1.NemoEntitystoreList{},
+	}
+
+	// Iterate over each Nemo microservice type and clean up disabled instances
+	for serviceType, serviceList := range serviceLists {
+		if err := r.List(ctx, serviceList, client.InNamespace(nemoService.Namespace)); err != nil {
```
varunrsekar, line 361: Instead of checking owner references later, we should filter here to only list objects owned by the parent CR (its a good practice to reduce memory footprint)
```suggestion
		if err := r.List(ctx, serviceList, client.InNamespace(nemoService.Namespace), client.MatchingLabels{<key>: <value>}); err != nil {
```
Instead of labels, we could also match owner references, but it would require setting up an index for it.

internal/controller/nemoservice_controller.go
```diff
@@ -0,0 +1,586 @@
...
+			if enabled, exists := enabledServices[serviceName]; !exists || !enabled {
+				continue
+			}
+
+			foundServices[serviceName] = true
+			state := getServiceState(svc)
+			serviceStates[serviceName] = state
+
+			switch state {
+			case appsv1alpha1.NemoServiceStatusFailed:
```
varunrsekar, line 482: how are we translating from a component CR state failed-> `NemoServiceStatusFailed` here? Its not reliable to assume the string value between them are the same.

api/apps/v1alpha1/nemoservice_types.go
```diff
@@ -0,0 +1,132 @@
...
+	State string `json:"state,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.state`,priority=0
+// +kubebuilder:printcolumn:name="Age",type="date",format="date-time",JSONPath=".metadata.creationTimestamp",priority=0
+
+// NemoService is the Schema for the nemoservices API
+type NemoService struct {
```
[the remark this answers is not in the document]
shivamerla: we can discuss on the naming

api/apps/v1alpha1/nemoservice_types.go
```diff
@@ -0,0 +1,132 @@
...
+	// Guardrails configures attributes to deploy a Nemo Guardrails service
+	Guardrails NemoServiceGuardrailsSpec `json:"guardrails,omitempty"`
+	// Entitystore configures attributes to deploy a Nemo Entitystore service
+	Entitystore NemoServiceEntitystoreSpec `json:"entitystore,omitempty"`
+}
+
+// NemoServiceCustomizerSpec defines the desired state of NemoCustomizer as part of the NeMoServices
+type NemoServiceCustomizerSpec struct {
+	Name    string              `json:"name,omitempty"`
+	Enabled *bool               `json:"enabled,omitempty"`
```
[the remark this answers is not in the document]
shivamerla: This field is to dynamically enable/disable services on the fly. Without users having to remove the entire config. By default we assume they are disabled.

shivamerla:
closing this as we were hitting the CRD size limit of 256KB when trying to deploy all services together. This needs to be reworked.

PR #338, opened 2025-02-21
Fixing DB env var for NeMo Services

visheshtanksale:


internal/controller/nemo_evaluator_controller.go
```diff
@@ -447,3 +460,21 @@ func (r *NemoEvaluatorReconciler) renderAndSyncResource(ctx context.Context, Nem
 	}
 	return nil
 }
+
+func (r *NemoEvaluatorReconciler) getKubernetesSecret(ctx context.Context, namespace, secretName, key string) (string, error) {
```
shivamerla, line 464: nit: getValueFromSecret() better suits this function

api/apps/v1alpha1/nemo_evaluator_types.go
```diff
@@ -827,6 +835,27 @@ func (n *NemoEvaluator) GetServiceMonitorAnnotations() map[string]string {
 	return NemoEvaluatorAnnotations
 }
 
+func (n *NemoEvaluator) GetSecretParams(secretValue string) *rendertypes.SecretParams {
```
shivamerla, line 838: better to get input param as map[string][string] and populate all key-value pairs in the data field.

internal/controller/nemocustomizer_controller_test.go
```diff
@@ -304,13 +318,22 @@ var _ = Describe("NemoCustomizer Controller", func() {
 		if err == nil {
 			Expect(k8sClient.Delete(ctx, resource)).To(Succeed())
 		}
+
+		namespacedName = types.NamespacedName{Name: "eval-pg-existing-secret", Namespace: "default"}
```
shivamerla, line 322: nit `ncs-pg-existing-secret`

internal/controller/nemocustomizer_controller_test.go
```diff
@@ -453,8 +476,15 @@ var _ = Describe("NemoCustomizer Controller", func() {
 			}))
 
 			Expect(envVars).To(ContainElement(corev1.EnvVar{
-				Name:  "POSTGRES_DB_DSN",
-				Value: nemoCustomizer.GeneratePostgresConnString(),
+				Name: "POSTGRES_DB_DSN",
```
shivamerla, line 479: we have to check with ncs team to align on the same env variable name for this.

shivamerla (approved):
LGTM!

internal/controller/nemocustomizer_controller_test.go
visheshtanksale: Fixed

api/apps/v1alpha1/nemo_evaluator_types.go
visheshtanksale: Fixed

internal/controller/nemo_evaluator_controller.go
visheshtanksale: Fixed

internal/controller/nemocustomizer_controller_test.go
visheshtanksale: It is `POSTGRES_DB_DSN` in the latest deployment.yaml. Do we need to check with them further?

shivamerla: i mean between all microservices, they need to be consistent. currently its POSTGRES_DB_URI and POSTGRES_DB_DSN. we have to suggest them.

shivamerla (approved):

PR #339, opened 2025-02-21
NeMo Dependencies Readme Update

angudadevops:


shivamerla:
@angudadevops can you please sign these commits. Also, can you remove the stale README files under each role.

angudadevops:
@shivamerla removed `READEMEs` under each role and I can see commits are signed and verified. let me know

PR #341, opened 2025-02-21
chore: Add rbac to allow guardrail to use anyuid scc in openshift

slu2011:
Signed-off-by: Shi Lu <shil@nvidia.com>


slu2011:
Will work with guardrails team to release a new guardrail container with non-root user builtin. This change is not needed after that.

PR #345, opened 2025-02-25
Remove mongodb from evaluator ansible install

visheshtanksale:


test/e2e/nemo-dependencies/values.yaml
```diff
@@ -4,7 +4,7 @@ install:
   embedding: no
   ranking: no
   entity_store: no
-  evaluator: no
+  evaluator: yes
```
shivamerla, line 7: lets leave this off by default and users enable necessary components.

visheshtanksale: Yes. Forgot to change it back after testing. Fixed it

shivamerla (approved):

PR #346, opened 2025-02-25
Update NIMService status with model information

varunrsekar:


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -87,11 +87,25 @@ type NIMCacheVolSpec struct {
 	Profile string `json:"profile,omitempty"`
 }
 
+type ModelEngine string
+
+const (
+	ModelEngineNIM ModelEngine = "nim"
```
shivamerla, line 93: NIM is not a model engine type. its either `vllm` or `trtllm`.  This will be misleading to add this in the status.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -275,6 +283,55 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
...
+	logger := log.FromContext(ctx)
+
+	// Lookup NIMCache instance in the same namespace as the NIMService instance
+	svc := &corev1.Service{}
+	if err := r.Get(ctx, types.NamespacedName{Name: nimService.GetName(), Namespace: nimService.GetNamespace()}, svc); err != nil {
+		logger.Error(err, "unable to fetch k8s service", "nimservice", nimService.GetName())
+		return "", err
+	}
+
+	return fmt.Sprintf("%s:%v", svc.Spec.ClusterIP, nimService.GetServicePort()), nil
```
shivamerla, line 332: For other service types e.g. `LoadBalancer`, this will be internal IP and the `EndpointAddress` in the status can be misleading. We have to make it explicit in that case that this is only for in-cluster access.

varunrsekar: I'll set ClusterIP and LoadBalancerIP in the status so we'd have both Internal and External endpoints to reach the service.

api/apps/v1alpha1/nimservice_types.go
shivamerla: This is something we should derive automatically when setting up Guardrail config rather than adding as NIM service status.

internal/controller/platform/standalone/nimservice.go
varunrsekar: Would Ingress IP/hostname take precedence over LoadBalancerIP from the service? Do you have any setups that I can take a look at? And I'm going to assume that if we have multiple ingresses, we'd only want to show one.

shivamerla: Yes, we should only check if ingress was enabled in the `NIMService` configuration, it would be like below.

```
ingress:
  enabled: true
  spec:
    ingressClassName: nginx
    rules:
      - host: demo.nvidia.example.com
        http:
          paths:
          - backend:
              service:
                name: meta-llama3-8b-instruct
                port:
                  number: 8000
            path: /v1/chat/completions
            pathType: Prefix
```

api/apps/v1alpha1/nimservice_types.go
varunrsekar: Removed model engine from the status

internal/controller/platform/standalone/nimservice.go
varunrsekar: Thanks. Added ingress host as model external endpoint if enabled.

shivamerla (approved):
LGTM

PR #348, opened 2025-02-27
Fix PVC creation with an empty storage class  name

shivamerla:


shengnuo:
LGTM

internal/shared/pvc.go
```diff
@@ -41,7 +40,12 @@ func ConstructPVC(pvc appsv1alpha1.PersistentVolumeClaim, pvcMeta metav1.ObjectM
 					corev1.ResourceStorage: size,
 				},
 			},
-			StorageClassName: &storageClassName,
 		},
-	}, nil
+	}
+
+	if pvc.StorageClass != "" {
```
varunrsekar, line 46: QQ: why not make `pvc.StorageClass` an optional string and pass it along as is?
```
// PersistentVolumeClaim defines the attributes of PVC used as a source for caching NIM model
type PersistentVolumeClaim struct {
...
        // StorageClass to be used for PVC creation. Leave it as empty if the PVC is already created.
-	StorageClass string `json:"storageClass,omitempty"`
+	StorageClass *string `json:"storageClass,omitempty"`
...
}
```

visheshtanksale (approved):
LGTM!!

mkhaas (commented):
LGTM

PR #354, opened 2025-03-04
Bump NeMo datastore version

shivamerla:


api/apps/v1alpha1/nemo_datastore_types.go
```diff
@@ -591,7 +591,7 @@ func (n *NemoDatastore) GetDefaultLivenessProbe() *corev1.Probe {
 		FailureThreshold:    3,
 		ProbeHandler: corev1.ProbeHandler{
 			HTTPGet: &corev1.HTTPGetAction{
-				Path: "/v1/health/live",
```
shengnuo, line 594: would we have to worry about backward compatibility?
If someone uses an older image, the probe endpoint would not work.
Should we make the probe endpoints configurable in the CR as well?

shengnuo:
LGTM

shengnuo (approved):
LGTM

PR #360, opened 2025-03-06
Fix model info in nimservice status

varunrsekar:
Changes:
* Fix model name detection for non-llm NIMs
* Don't throw error if model name detection fails
* Use Ingress loadbalancerIP as `externalEndpoint` if no host is specified in ingress spec rules.


internal/controller/platform/standalone/nimservice.go
```diff
@@ -302,24 +302,31 @@ func (r *NIMServiceReconciler) updateModelStatus(ctx context.Context, nimService
...
+			return modelsList.Data[0].Id, nil
+		}
+
 		for _, model := range modelsList.Data {
 			if model.Object != nimmodels.ObjectTypeModel {
 				continue
 			}
-			if model.Root == nil || *model.Root != model.Id {
-				continue
+			if model.Root != nil && *model.Root == model.Id {
```
visheshtanksale, line 322: Whats the significance of `model.Root` ? Cant we just return id?

varunrsekar: When we have lora adapters, we need `model.Root` to identify the model.
https://github.com/NVIDIA/k8s-nim-operator/blob/0f5bda89cf715d58159c5213d06d893036564ea3/internal/controller/platform/standalone/nimservice_test.go#L843
^ this has a crude example of what the `/v1/models` API response looks like.

shivamerla (approved):
LGTM

internal/controller/platform/standalone/nimservice.go
visheshtanksale: OK. Thanks for clarifying that.

visheshtanksale (approved):
LGTM

PR #362, opened 2025-03-07
Add image pull secrets to nemodatastore CR sample and remove authSecret from NeMoDatastore Spec

shengnuo:


config/samples/apps_v1alpha1_nemodatastore.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   name: nemodatastore-sample
   namespace: nemo
 spec:
-  authSecret: ngc-secret
+  authSecret: ngc-auth-secret
```
shivamerla, line 10: not able to recollect if the auth-secret is needed for datastore or not, can you confirm?

shivamerla (approved):
LGTM

PR #363, opened 2025-03-07
Update CR samples with defaults from ansible playbook

shivamerla:


config/samples/apps_v1alpha1_nimservice.yaml
```diff
@@ -7,8 +7,8 @@ metadata:
   name: meta-llama3-8b-instruct
 spec:
   image:
-    repository: nvcr.io/nim/meta/llama3-8b-instruct
-    tag: 1.0.3
+    repository: nvcr.io/nim/meta/llama-3.1-8b-instruct:
```
varunrsekar, line 10: nit: typo?
```suggestion
    repository: nvcr.io/nim/meta/llama-3.1-8b-instruct
```

config/samples/apps_v1alpha1_nemodatastore.yaml
```diff
@@ -19,7 +19,7 @@ spec:
       secretName: nds-minio-existing-secret
```
varunrsekar, line 19: is this supposed to `datastore-minio-*` as well?

varunrsekar (changes requested):

config/samples/apps_v1alpha1_nemocustomizer.yaml
```diff
@@ -9,15 +9,16 @@ metadata:
 spec:
   wandbSecret:
     name: wandb-secret
-    apiKeyKey: encryption_key
+    apiKeyKey: encryptionKey
```
shengnuo, line 12: I might not have the most update to date doc, but the NMP helm template is still defaulting the wandb api key in secret as `encryption_key`
https://gitlab-master.nvidia.com/aire/microservices/nmp/-/blob/main/helm/platform/components/customizer/templates/_helpers.tpl#L144

shivamerla: This is to align with the default `key` that we have used in the ansible playbook [here](https://github.com/NVIDIA/k8s-nim-operator/blob/main/test/e2e/nemo-dependencies/customizer/templates/secrets.yaml.j2#L19)

config/samples/apps_v1alpha1_nemodatastore.yaml
shivamerla: we need to update ansible playbook, but this is based on the current secret name we are using [here](https://github.com/NVIDIA/k8s-nim-operator/blob/main/test/e2e/nemo-dependencies/datastore/templates/secrets.yaml.j2#L351).

varunrsekar: Thanks. can you update that template? The helm_release_name is `datastore-minio`, so its weird for the secret to be `nds-minio*` (similarly for `nds-pg`)

shivamerla: yes, was thinking same. I have updated ansible templates as well.

varunrsekar (approved):
LGTM

PR #365, opened 2025-03-10
Retrieve NIM_ENDPOINT_API_KEY for Guardrail from k8s secret

shengnuo:
This MR injects the `NIM_ENDPOINT_API_KEY` environment variable into the NeMo Guardrail container from a Kubernetes Secret.

Before creating the Guardrail service, a Secret with key `NIM_ENDPOINT_API_KEY` must be created. 

`.spec.authSecret` in NemoGuardrail is now optional, as it's only required when the NIM is NGC hosted.

### DIFF
--- a/api/apps/v1alpha1/nemo_guardrails_types.go
+++ b/api/apps/v1alpha1/nemo_guardrails_types.go
@@ -59,8 +59,8 @@ type NemoGuardrailSpec struct {
 	Command []string        `json:"command,omitempty"`
 	Args    []string        `json:"args,omitempty"`
 	Env     []corev1.EnvVar `json:"env,omitempty"`
-	// The name of an secret that contains authn for the NGC NIM service API
-	AuthSecret string `json:"authSecret"`
+	// The name of an secret that contains authn for the NGC NIM service API, required when the NIM is hosted by NGC
+	AuthSecret string `json:"authSecret,omitempty"`
 	// ConfigStore stores the config of the guardrail service
 	ConfigStore    GuardrailConfig              `json:"configStore,omitempty"`
 	Labels         map[string]string            `json:"labels,omitempty"`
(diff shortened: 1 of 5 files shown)


api/apps/v1alpha1/nemo_guardrails_types.go
```diff
@@ -187,6 +187,17 @@ func (n *NemoGuardrail) GetStandardEnv() []corev1.EnvVar {
 			Name:  "DEMO",
 			Value: "False",
 		},
+		{
+			Name: "NIM_ENDPOINT_API_KEY",
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					Key: "NIM_ENDPOINT_API_KEY",
```
varunrsekar, line 194: Per the [documentation](https://developer.nvidia.com/docs/nemo-microservices/guardrails/source/getting-started/deploy-helm.html) on Guardrails, this key is `nim-endpoint-api-key`. Should this key be a field in the CR instead?

shengnuo: ya I agree, having a customizable key field would be better for user experience.
As the similar `authSecret` is used by other NIM services, I will handle that in a separate MR

api/apps/v1alpha1/nemo_guardrails_types.go
```diff
@@ -187,6 +187,17 @@ func (n *NemoGuardrail) GetStandardEnv() []corev1.EnvVar {
 			Name:  "DEMO",
 			Value: "False",
 		},
+		{
```
shivamerla, line 190: the `authSecret` is optional and used only when NIMs are hosted by NGC. This should be added conditionally as well.

shengnuo: `authSecret` is currently a required field for GuardRail 
https://github.com/NVIDIA/k8s-nim-operator/blob/311fc567c6d6262bcdebee9832c5bd44d9911074/config/crd/bases/apps.nvidia.com_nemoguardrails.yaml#L2207

We should make it optional if it's not always required

shengnuo: `.spec.authSecret` in NemoGuardrail is now optional

api/apps/v1alpha1/nemo_guardrails_types.go
```diff
@@ -57,8 +57,8 @@ type NemoGuardrailSpec struct {
 	Command []string        `json:"command,omitempty"`
 	Args    []string        `json:"args,omitempty"`
 	Env     []corev1.EnvVar `json:"env,omitempty"`
-	// The name of an secret that contains authn for the NGC NIM service API
-	AuthSecret string `json:"authSecret"`
+	// The name of an secret that contains authn for the NGC NIM service API, required when the NIM is hosted by NGC
+	AuthSecret *string `json:"authSecret,omitempty"`
```
shivamerla, line 61: Pointer type is not required here. Pointer type is useful when we need to distinguish between unset (nil) and empty values (""), this is not the case here. Empty value have no special meaning in this case.

shivamerla (approved):

shivamerla (approved):

shivamerla (approved):

PR #366, opened 2025-03-11
Fix Milvus installation in OpenShift

shengnuo:
Milvus is crashing when it's installed in a custom namespace OpenShift with the following error message
`2025/02/27 07:46:13 write failed: open /milvus/configs/milvus.yaml: permission denied`

This stems from the `restricted-v2` SCC in OpenShift. 
By default, OpenShift uses the most restricted SCC `restricted-v2` for service accounts in `nemo` namespace. When such SCC is applied, the following securityContext is added to the Milvus container.
```yaml
securityContext:
allowPrivilegeEscalation: false
capabilities:
drop:
- ALL
runAsNonRoot: true
runAsUser: 1000790000
```

This makes the main container unable to read the `/milvus/configs/milvus.yaml` file created by the init container.

This MR fixes the issue by creating a dedicated service account for Milvus, and assigning the SCC `anyuid` to the service account.

### DIFF
--- a/test/e2e/nemo-dependencies/evaluator/tasks/milvus.yaml
+++ b/test/e2e/nemo-dependencies/evaluator/tasks/milvus.yaml
@@ -1,4 +1,20 @@
 ---
+- name: Get Kube API resources
+  command: kubectl api-resources --verbs=list --namespaced -o name
+  register: api_resources
+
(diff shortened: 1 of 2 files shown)


shivamerla (approved):
LGTM

PR #367, opened 2025-03-12
Nemo Guardrails CR fixes

varunrsekar:
* fix default probe endpoints
* create guardrails configstore PVC if requested
* abstract out configstore configMap into a separate object and validate its not provided along with PVC.


internal/controller/nemo_guardrail_controller.go
```diff
@@ -394,6 +400,41 @@ func (r *NemoGuardrailReconciler) reconcileNemoGuardrail(ctx context.Context, Ne
...
+		if nemoGuardrail.Spec.ConfigStore.PVC.Create != nil && *nemoGuardrail.Spec.ConfigStore.PVC.Create {
+			pvc, err = shared.ConstructPVC(*nemoGuardrail.Spec.ConfigStore.PVC, metav1.ObjectMeta{Name: pvcName, Namespace: nemoGuardrail.GetNamespace()})
+			if err != nil {
+				logger.Error(err, "Failed to construct pvc", "name", pvcName)
+				return err
+			}
+			if err := controllerutil.SetControllerReference(nemoGuardrail, pvc, r.GetScheme()); err != nil {
+				return err
+			}
+			err = r.Create(ctx, pvc)
```
shivamerla, line 424: Will guardrail service assume a `default` config with an empty PVC?

api/apps/v1alpha1/nemo_guardrails_types.go
```diff
@@ -81,11 +81,19 @@ type NemoGuardrailSpec struct {
 	RuntimeClass string `json:"runtimeClass,omitempty"`
 }
 
+// GuardrailConfig defines the source where the service config is made available.
+//
+// +kubebuilder:validation:XValidation:rule="!(has(self.configMap) && has(self.pvc))", message="Cannot set both ConfigMap and PVC in ConfigStore"
 type GuardrailConfig struct {
-	ConfigMap string                 `json:"configMap,omitempty"`
+	ConfigMap *ConfigMap             `json:"configMap,omitempty"`
```
shivamerla, line 88: ConfigMap was used for testing purposes and can be removed. Guardrail config can be updated dynamically through the API and those will be lost if we are using a CM. @slu2011 thoughts?

internal/controller/nemo_guardrail_controller.go
varunrsekar: Guardrail service can start with no config defined if that's what you're asking. Invoking `/v1/guardrail/configs` returns an empty list:
```
$ curl http://nemo-guardrails.nemo-guardrails.svc.cluster.local:7331/v1/guardrail/configs
{"object":"list","data":[],"pagination":{"page":1,"page_size":10,"current_page_size":0,"total_pages":0,"total_results":0},"sort":"created_at"}
```

api/apps/v1alpha1/nemo_guardrails_types.go
varunrsekar: I dont mind either way. Although the CM makes for a quick way to test the CR since it comes pre-populated with a config.

shivamerla: Ok, since CM is always mounted as `read-only` any POST call to update the config should return an error and we can document that clearly.

shivamerla (approved):

PR #374, opened 2025-03-18
Cleanup samples

varunrsekar:


config/samples/apps_v1alpha1_nemocustomizer.yaml
```diff
@@ -1,9 +1,6 @@
 apiVersion: apps.nvidia.com/v1alpha1
 kind: NemoCustomizer
 metadata:
-  labels:
-    app.kubernetes.io/name: k8s-nim-operator
```
shivamerla, line 5: Lets keep this label, its easy for querying all CR statuses with a single command.

varunrsekar: So it'll be confusing when all its child entities have the following set of labels:
```
    app.kubernetes.io/instance: nemocustomizer-sample
    app.kubernetes.io/managed-by: k8s-nim-operator
    app.kubernetes.io/name: nemocustomizer-sample
    app.kubernetes.io/operator-version: ""
    app.kubernetes.io/part-of: nemo-customizer-service
```
while the parent CR has `app.kubernetes.io/name: k8s-nim-operator`.

Ideally `part-of` label for child entities would match the `name` label of the parent entity.

varunrsekar: > Lets keep this label, its easy for querying all CR statuses with a single command.

Also didnt quite understand this. What is the single command for getting all CR statuses?

PR #376, opened 2025-03-19
Remove e2e test workflow

elezar:
Removing the e2e test workflow since this does not provide a reasonable signal of health.

The e2e tests always fail (https://github.com/NVIDIA/k8s-nim-operator/actions/workflows/e2e.yml) and do not provide a reasonable signal of health. If these are readded, they should be configured so as to run per PR.


.github/workflows/e2e.yml
```diff
@@ -1,91 +0,0 @@
...
-      - completed
-    branches:
-      - "pull-request/[0-9]+"
-      - main
-      - release-*
-
-jobs:
-  e2e-tests:
-    runs-on: linux-amd64-cpu4
-    if: ${{ github.event.workflow_run.conclusion == 'success' }} && ${{ github.event.workflow_run.event == 'push' }}
```
varunrsekar, line 30:
```
if: false # TODO: Disabled until e2e tests is made stable
```
Can you instead comment this line and replace with the above?

varunrsekar (changes requested):

shivamerla:
thanks @elezar we didn't get to check the errors with EC2 provisioning. @ArangoGutierrez any idea why `apt update` is failing during provisioning using Holodeck? https://github.com/NVIDIA/k8s-nim-operator/actions/runs/13848216533/job/38750735288

shivamerla:
closing this as this is disabled with https://github.com/NVIDIA/k8s-nim-operator/pull/411

PR #378, opened 2025-03-20
Rework object hash annotations

varunrsekar:


internal/utils/utils.go
```diff
@@ -293,3 +290,109 @@ func FormatEndpoint(ip string, port int32) string {
 	}
 	return fmt.Sprintf("%s:%d", ip, port)
 }
+
+func DeepHashObject(objToWrite any) string {
+	hasher := fnv.New32a()
+	hasher.Reset()
+	fmt.Fprintf(hasher, "%v", dump.ForHash(objToWrite))
+	return rand.SafeEncodeString(fmt.Sprint(hasher.Sum32()))
+}
```
shivamerla, line 299: This doesn't return a consistent hash when the order of elements change. Hence we had to sort them earlier.

internal/utils/utils.go
```diff
@@ -293,3 +290,109 @@ func FormatEndpoint(ip string, port int32) string {
...
+
+func UpdateObject(obj client.Object, desired client.Object) client.Object {
+	if obj == nil || desired == nil || !reflect.DeepEqual(obj.GetObjectKind(), desired.GetObjectKind()) || obj.GetName() != desired.GetName() || obj.GetNamespace() != desired.GetNamespace() {
+		panic("invalid input to UpdateObject")
+	}
+
+	switch castedObj := obj.(type) {
```
shivamerla, line 306: We would need to add `ConfigMap` and `Secret` object as well.

varunrsekar: this first draft was just for nimservice. Will add those once I get to the rest of the CRs

internal/controller/platform/standalone/nimservice.go
```diff
@@ -385,6 +385,16 @@ func (r *NIMServiceReconciler) renderAndSyncResource(ctx context.Context, nimSer
 
 	namespacedName := types.NamespacedName{Name: nimService.GetName(), Namespace: nimService.GetNamespace()}
 
+	err := r.Get(ctx, namespacedName, obj)
+	if err != nil && !errors.IsNotFound(err) {
+		logger.Error(err, fmt.Sprintf("Error is not NotFound for %s: %v", obj.GetObjectKind(), err))
+		return err
+	}
+	// Don't do anything if CR is unchanged.
+	if err == nil && !utils.IsParentSpecChanged(obj, utils.DeepHashObject(nimService.Spec)) {
```
varunrsekar, line 394: For deployments/statefulsets, also check contents of dependency secrets/configmaps

varunrsekar: We shouldn't really reach into contents of configmaps/secrets not controlled by nim-operator. Dropping this issue

shivamerla (approved):
LGTM! thanks @varunrsekar

PR #381, opened 2025-03-24
Replacing kubectl with crd-apply-tool

visheshtanksale:
- Remove kubectl to reduce CVEs
- Replacing it with [crd apply tool](https://github.com/NVIDIA/k8s-operator-libs/tree/main/pkg/crdutil)


shivamerla (approved):
LGTM!

PR #384, opened 2025-03-31
Add versioned samples and symlink to latest nemo samples

shivamerla:


config/samples/nemo/latest
```diff
@@ -0,0 +1 @@
+25.03
```
varunrsekar, line 1: Can we rename the `25.03` folder to `latest` and this file to `VERSION` inside it?
I'm not sure why this needs to be symbolic link with the version as contents.

shengnuo:
Seems like we are going down the path of maintaining different NeMo CRs for different NeMo microservice versions.
Since 25.03 CRs will only work with a specific version of NIM operator, should we specify NIM Operator version to use in the CR directory?

config/samples/nemo/latest
shivamerla: this is how symbolic link is stored in Git, the link itself is named as `latest` and contains the target path here.

varunrsekar: No I mean, its confusing looking it up on github: https://github.com/shivamerla/k8s-nim-operator/blob/versioned_samples/config/samples/nemo/latest/apps_v1alpha1_nemodatastore.yaml
^ This wouldnt go to that 25.03 sample.

shivamerla: ah i didn't mean to use it that way but from ansible scripts etc to deploy latest ones. But now i think more about it, it makes sense to always keep `latest` folder and link in the docs. For older versions, we keep moving to version specific folders.

shengnuo:
LGTM

shengnuo (approved):

mkhaas:
LGTM

varunrsekar (approved):
LGTM

config/samples/nemo/latest/README
```diff
@@ -0,0 +1,22 @@
+# NeMo Custom Resources for 25.04
```
varunrsekar, line 1: Let’s not mention the version here. This can very easily go stale

varunrsekar (approved):

config/samples/nemo/latest/README
shivamerla: we can cleanup when we update with 25.04 bits.

PR #390, opened 2025-04-01
Update guardrail sample to use PVC as a config store and cleanup ansible

shivamerla:


config/samples/nemo/latest/apps_v1alpha1_nemoguardrails.yaml
```diff
@@ -6,8 +6,12 @@ metadata:
 spec:
   authSecret: ngc-api-secret
   configStore:
-    configMap:
-      name: gr-config
+    pvc:
+      name: "pvc-guardrail-config"
+      create: true
+      storageClass: ""
```
varunrsekar, line 12:
```
  storageClass	<string>
    StorageClass to be used for PVC creation. Leave it as empty if the PVC is
    already created.
```
Is this meant to be empty? You'd end up with a `FailedBinding` error in this case

varunrsekar (changes requested):

config/samples/nemo/latest/apps_v1alpha1_nemoguardrails.yaml
shivamerla: that comment needs to be updated. Instead it should say `Leave it as empty if a default storage class is already provisioned`. Also, if we set it as empty here, then in the code we don't pass this at all. This was fixed.

config/samples/nemo/latest/apps_v1alpha1_nemodatastore.yaml
```diff
@@ -31,7 +31,7 @@ spec:
   pvc:
     name: "pvc-shared-data"
     create: true
-    storageClass: "local-path"
+    storageClass: ""
```
shengnuo, line 34: since the storageClass is optional, maybe we can leave it out of the CR.
I worry that this might confuse the users into believing the resulted PVC will have an empty string for the `.spec.storageClassName` field.

shengnuo: or maybe use a placeholder for storageClass
```
storageClass: <my-storage-class>
```

shivamerla: In the docs, thats how they are, but the samples we have are configured with defaults from ansible. There are cases for e.g. TKGS where a default SC is already there and causing issues with the default sample.

config/samples/nemo/latest/apps_v1alpha1_nemoguardrails.yaml
varunrsekar: hmm. I just tried it out. Even though we dont set a storageClass in the PVC spec, it doesnt help if the cluster doesnt have a default class:
```
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi
  volumeMode: Filesystem
```
Error:
```
Events:
  Type    Reason         Age               From                         Message
  ----    ------         ----              ----                         -------
  Normal  FailedBinding  1s (x3 over 16s)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
```

shivamerla: https://github.com/NVIDIA/k8s-nim-operator/pull/391

shivamerla: description is updated now with the storageclass field. In the docs this is called out as well.

shivamerla: no, we don't want to cover all pre-requisites there. Here are the use cases

1. User deploys all dependencies through ansible (which will set the "local-path" sc as default). In this case samples work readily.
2. User deploys all dependencies through ansible but not local-path-provisioner. In that case, they need to explicitly disable that through ansible where we have a comment about requirement to have a default-storage class in the cluster.
3. User manages all dependencies and savvy to edit CR samples with all custom values (dependent service endpoints, secrets, storage classes etc)

varunrsekar: Gotcha. This makes sense

config/samples/nemo/latest/apps_v1alpha1_nemoguardrails.yaml
```diff
@@ -4,10 +4,15 @@ metadata:
   name: nemoguardrails-sample
   namespace: nemo
 spec:
-  authSecret: ngc-api-secret
```
varunrsekar, line 7: QQ: Is this required?

Per docs, I believe `ngc-api-secret` only sets `NGC_API_KEY`. Guardrail needs `NIM_ENDPOINT_API_KEY` in the `spec.authSecret`

a-mccarthy: I think we do need to update this in the docs. What would be the value for NIM_ENDPOINT_API_KEY?

shivamerla: @varunrsekar @a-mccarthy this is required only when the NIM that we are safeguarding is hosted by NVIDIA (i.e ai.nvidia.com). For the operator use case NIMs are deployed locally in the cluster and hence we can remove this by default from the sample. For documentation we still need to cover this param explaining this secret have to contain key `NIM_ENDPOINT_API_KEY` and value as token, which will be imported as an env in the container.

test/e2e/nemo-dependencies/customizer/templates/pvc.yaml.j2
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: finetuning-ms-models-pvc
```
varunrsekar, line 4: Thinking on this:
I'm wondering if this PVC needs to be an input in the customizer CR as `spec.modelCache`. Would it make sense for the PVC lifecycle to be independent of the customizer or tightly coupled to it?

shivamerla: resolving based on offline discussion on this.

PR #392, opened 2025-04-01
Add an ansible task to create a PVC for base model storage for finetuning

shivamerla:
…ning

### DIFF
--- a/test/e2e/nemo-dependencies/customizer/templates/pvc.yaml.j2
+++ b/test/e2e/nemo-dependencies/customizer/templates/pvc.yaml.j2
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: finetuning-ms-models-pvc
+  namespace: {{ namespace }}
+spec:
+  storageClassName: {{ '"local-path"' if localPathProvisioner.enabled else '""' }}
+  accessModes:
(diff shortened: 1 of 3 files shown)


test/e2e/nemo-dependencies/customizer/templates/pvc.yaml.j2
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: finetuning-ms-models-pvc
+  namespace: {{ namespace }}
+spec:
+  storageClassName: {{ '"local-path"' if localPathProvisioner.enabled else '""' }}
+  accessModes:
+    - ReadWriteOnce
```
varunrsekar, line 9: This is RWO for local-path but otherwise this can be RWM for NFS etc.

shivamerla: Yes, there is no easy way to check if the default provisioner (non local-path) supports RWX volumes. It is safe to default to RWO and document to create a RWX volume otherwise.

shivamerla: I have folded this MR into https://github.com/NVIDIA/k8s-nim-operator/pull/390 and added options to provide RWX access mode.

PR #397, opened 2025-04-03
Update values.yaml

mkhaas:
Changing default to yes

### DIFF
--- a/test/e2e/nemo-dependencies/values.yaml
+++ b/test/e2e/nemo-dependencies/values.yaml
@@ -1,8 +1,8 @@
 install:
-  customizer: no
-  datastore: no
-  entity_store: no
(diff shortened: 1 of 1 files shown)


shivamerla (approved):

test/e2e/nemo-dependencies/values.yaml
```diff
@@ -1,8 +1,8 @@
...
-  customizer: no
-  datastore: no
-  entity_store: no
-  evaluator: no
+  customizer: yes
+  datastore: yes
+  entity_store: yes
+  evaluator: yes
 
 uninstall:
```
varunrsekar, line 7: should uninstall be defaulted to `yes` as well?

PR #408, opened 2025-04-03
enable atleast 1 model in cust config sample

varunrsekar:


shivamerla (approved):
thanks @varunrsekar

mkhaas (commented):
I think we should only enable one model -- not all -   Llama 3.2 1B Instruct model. -- since all these models get downloaded to the PVC.

varunrsekar:
> I think we should only enable one model -- not all - Llama 3.2 1B Instruct model. -- since all these models get downloaded to the PVC.

@mkhaas NMP helmchart has all these 4 models enabled by default

shivamerla (approved):

PR #429, opened 2025-04-15
Adding support for proxy to NIM Service

visheshtanksale:
- User needs to add the CA cert used by TLS proxy in a config map
- Provide CA cert config map in the `ProxySpec` on NIM Service object
- Specify the HTTP and HTTPS proxy endpoints on `ProxySpec`

Working on adding Unit tests


mkhaas (commented):
lgtm

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -480,6 +536,37 @@ func (n *NIMService) GetDeploymentKind() string {
 	return "Deployment"
 }
 
+// GetInitContainers returns the init containers for the NIMService deployment
+func (n *NIMService) GetInitContainers() []corev1.Container {
+	if n.Spec.Proxy != nil {
+		return []corev1.Container{
```
shivamerla, line 542: Wondering if we should move the generation of initContainer spec for proxy into a k8sUtil package. That way we can leverage same for NIMCache, and other NeMo services that are pulling models from NGC/HF/S3 etc.

visheshtanksale: Cannot move the method because image of the container is based on the specific object like NIMCache, NIMService.

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -78,10 +78,18 @@ type NIMServiceSpec struct {
...
-	GroupID          *int64 `json:"groupID,omitempty"`
-	RuntimeClassName string `json:"runtimeClassName,omitempty"`
+	Replicas         int        `json:"replicas,omitempty"`
+	UserID           *int64     `json:"userID,omitempty"`
+	GroupID          *int64     `json:"groupID,omitempty"`
+	RuntimeClassName string     `json:"runtimeClassName,omitempty"`
+	Proxy            *ProxySpec `json:"proxy,omitempty"`
+}
+
+// ProxySpec defines the proxy configuration for NIMService
```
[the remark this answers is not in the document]
visheshtanksale: Added

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -480,6 +536,37 @@ func (n *NIMService) GetDeploymentKind() string {
...
+	if n.Spec.Proxy != nil {
+		return []corev1.Container{
+			{
+				Name:            "update-ca-certificates",
+				Image:           n.GetImage(),
+				ImagePullPolicy: corev1.PullPolicy(n.GetImagePullPolicy()),
+				Command: []string{
+					"sh", "-c", "cp /custom/*.crt /usr/local/share/ca-certificates/ && update-ca-certificates --verbose && cp -r /etc/ssl/certs/ /ca-certs",
+				},
+				SecurityContext: &corev1.SecurityContext{
```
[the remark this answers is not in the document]
visheshtanksale: Is this managed based on the annotations?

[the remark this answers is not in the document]
shivamerla: no, we have to adjust the SCC rules as well.

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -417,6 +450,22 @@ func (n *NIMService) GetVolumes(modelPVC PersistentVolumeClaim) []corev1.Volume
 		},
 	}
 
+	if n.GetProxySpec() != nil {
```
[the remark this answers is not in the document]
shivamerla: @visheshtanksale for OCP, we have to update SCC rules to use `anyuid` SCC instead, also the annotation to use that SCC when proxy spec is set.

internal/controller/nimcache_controller.go
```diff
@@ -1194,14 +1192,33 @@ func (r *NIMCacheReconciler) constructJob(ctx context.Context, nimCache *appsv1a
...
+		if nimCache.GetProxySpec() != nil {
+			job.Spec.Template.Spec.InitContainers = nimCache.GetInitContainers()
+			job.Spec.Template.Spec.Containers[0].Env = utils.MergeEnvVars(job.Spec.Template.Spec.Containers[0].Env, nimCache.GetEnvWithProxy())
+			job.Spec.Template.Spec.Containers[0].VolumeMounts = append(job.Spec.Template.Spec.Containers[0].VolumeMounts,
+				corev1.VolumeMount{
+					Name:      "ca-cert-volume",
+					MountPath: "/etc/ssl",
+				},
+			)
+			job.Spec.Template.Spec.Volumes = append(job.Spec.Template.Spec.Volumes,
```
shivamerla, line 1208: we can move the spec for these volumes into a common method between nimcache and nimservice and `ca-cert-volume` as a shared constant.

internal/controller/nimcache_controller.go
```diff
@@ -959,9 +959,8 @@ func constructPodSpec(nimCache *appsv1alpha1.NIMCache, platformType k8sutil.Orch
 				},
 			},
 			SecurityContext: &corev1.PodSecurityContext{
-				RunAsUser:    nimCache.GetUserID(),
-				FSGroup:      nimCache.GetGroupID(),
-				RunAsNonRoot: ptr.To[bool](true),
```
shivamerla, line 964: Rather than removing it here, if proxy is set, then we can toggle this to `false` conditionally. That way the security config will remain same.

internal/controller/nimcache_controller.go
visheshtanksale: Updated this!

internal/controller/nimcache_controller.go
visheshtanksale: Actually, the container `RunAsNonRoot` overides the one on pod. So, made the change to set  `RunAsNonRoot: false` on init container

shivamerla:
Thanks @visheshtanksale changes look good. Once the SCC rules are updated, we can merge this.

shivamerla (approved):
LGTM

PR #437, opened 2025-04-17
Fix guardrail API

varunrsekar:
1. Make nimEndpoint optional to support multiple locally deployed NIMs through a global configstore config for guardrails
2. Fix mountpath when using a configmap for configstore so that its easier to use
3. Use a default `subPath` (guardrails-config-store) if `nemoguardrail.spec.configStore.pvc.subPath` is not specified so that any FS related files (like `lost+found` for ext4) are not present in the configstore path for guardrails.


shivamerla (approved):
LGTM

mkhaas (commented):
lgtm

PR #457, opened 2025-04-24
Update values.yaml

mkhaas:
Making it easier to run for local single node deployment. We will document in pre-req RWX is required for greater than 2 nodes.


test/e2e/nemo-dependencies/values.yaml
```diff
@@ -18,8 +18,8 @@ installation_namespace: nemo
 pvc:
   # Ignored when localPathProvisioner.enabled is true — in which case these will default to:
   # storage_class: "local-path" and volume_access_mode: "ReadWriteOnce"
-  storage_class: ""
```
shivamerla, line 21: @mkhaas these will be ignored and set to those exact values when local-path-provisioner is enabled.

PR #460, opened 2025-04-24
Make notebook writable

shivamerla:


test/e2e/nemo-dependencies/jupyter/templates/deployment.yaml.j2
```diff
@@ -13,6 +13,20 @@ spec:
...
     spec:
+      initContainers:
+        - name: copy-notebook-config
+          image: busybox
+          command:
+            - sh
+            - -c
+            - |
+              cp /config/{{ config_name }} /work/{{ config_name }}
+              cp /config/{{ notebook_name }} /work/{{ notebook_name }}
```
varunrsekar: Wouldnt this revert all notebook changes if the jupyter pod restarts?
Can you instead only run this if the files dont exist in the PVC?

shivamerla: good point, fixed it.

varunrsekar (approved):
LGTM

PR #479, opened 2025-05-06
Updating the golangci-lint to v2

visheshtanksale:
Since the github CI action has moved to golangci-lint v2 [here](https://github.com/NVIDIA/k8s-nim-operator/commit/c3408d666a52537eac5d288861d23b3abc66109e) , moving the make target to use v2 as well


Makefile
```diff
@@ -242,7 +242,7 @@ $(ENVTEST): $(LOCALBIN)
 .PHONY: golangci-lint
 golangci-lint: $(GOLANGCI_LINT) ## Download golangci-lint locally if necessary.
 $(GOLANGCI_LINT): $(LOCALBIN)
-	@GOBIN=$(PROJECT_DIR)/bin GO111MODULE=on $(GO_CMD) install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
+	@GOBIN=$(PROJECT_DIR)/bin GO111MODULE=on $(GO_CMD) install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
```
shivamerla, line 245: can we also bump GH action to `golangci/golangci-lint-action@v7`

visheshtanksale: @shivamerla Its already at `golangci/golangci-lint-action@v8`

shivamerla: Also, add a .golangci.yaml similar to [here](https://github.com/NVIDIA/k8s-dra-driver-gpu/blob/main/.golangci.yaml) for all linter checks.

shivamerla (approved):
Good to see all these linter fixes.

PR #485, opened 2025-05-09
Adding resources on upgrade crds

visheshtanksale:


deployments/helm/k8s-nim-operator/templates/upgrade_crd.yaml
```diff
@@ -83,16 +83,17 @@ spec:
...
-          imagePullPolicy: {{ .Values.operator.image.pullPolicy }}
-          securityContext:
-            allowPrivilegeEscalation: false
-            capabilities:
-              drop:
-              - ALL
-          command: ["crd-apply-tool"]
-          args:
-            - "--crds-dir=/opt/nim-operator/crds/"
+      - name: upgrade-crd
```
shivamerla, line 86: can we keep the same indentation here.

visheshtanksale: Fixed

shivamerla (approved):

PR #490, opened 2025-05-12
Add optional grpc,metrics ports to NIMService

varunrsekar:
This change introduces 2 new ports to the NIMService API spec:
* `spec.expose.service.grpcPort`: sets the application env `NIM_TRITION_GRPC_PORT`
* `spec.expose.service.metricsPort`: sets the application env `NIM_TRITON_METRICS_PORT`

### DIFF
--- a/api/apps/v1alpha1/nimservice_types.go
+++ b/api/apps/v1alpha1/nimservice_types.go
@@ -40,8 +40,6 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 const (
-	// NIMAPIPort is the default port that the NIM serves on.
-	NIMAPIPort = 8000
 	// NIMServiceConditionReady indicates that the NIM deployment is ready.
 	NIMServiceConditionReady = "NIM_SERVICE_READY"
 	// NIMServiceConditionFailed indicates that the NIM deployment has failed.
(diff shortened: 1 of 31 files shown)


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -881,6 +928,14 @@ func (n *NIMService) GetServiceMonitorParams() *rendertypes.ServiceMonitorParams
 			},
 		},
 	}
+	if n.Spec.Expose.Service.MetricsPort != nil {
+		smSpec.Endpoints = append(smSpec.Endpoints, monitoringv1.Endpoint{
+			Path:          "/metrics",
```
shivamerla, line 933: Shouldn't this be `v1/models`?

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -881,6 +928,14 @@ func (n *NIMService) GetServiceMonitorParams() *rendertypes.ServiceMonitorParams
 			},
 		},
 	}
+	if n.Spec.Expose.Service.MetricsPort != nil {
```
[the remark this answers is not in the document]
varunrsekar: For non-LLM NIMs:
* `/v1/metrics` on the API port is the application metrics - This includes `python_*`, `process_*` and `gpu_*` metrics. 
* And `/metrics` on the metrics port is the Triton Inference server metrics - There's some overlay with `/v1/metrics` but this includes a more comprehensive list of metrics

Where as for LLM NIMs:
Both framework metrics and the backend metrics are combined under the `/v1/metrics` endpoint.

For parity, we would need to expose all the metrics.

api/apps/v1alpha1/nimservice_types.go
varunrsekar: Why `/v1/models`? The Triton Metrics service only exposes the `/metrics` endpoint.

shivamerla: missed that triton is still reporting through `/metrics` endpoint. makes sense now.

shivamerla (approved):

PR #498, opened 2025-05-16
fix model status for NIMs incompatible with openapi schema

varunrsekar:
Fallback to using `/v1/metadata` API endpoint to get the model name for NIMs incompatible with openapi schema

The workflow:
* Try /v1/models API
* if it errors out with a 404 HTTP error, then use /v1/metadata API


shivamerla (approved):
LGTM

a-mccarthy:
@varunrsekar are there any customer impact of docs updates needed for this change?

PR #504, opened 2025-05-24
DRAFT: Fix nimsvc state

varunrsekar:
Changes:
* Retry nimservice reconcile every 5 seconds while nimcache is in progress
* Mark nimservice as failed if its nimcache is failed
* Retry nimservice reconcile if status updates hit api conflict errors

### DIFF
--- a/internal/controller/platform/standalone/standalone.go
+++ b/internal/controller/platform/standalone/standalone.go
@@ -115,6 +115,10 @@ func (s *Standalone) Sync(ctx context.Context, r shared.Reconciler, resource cli
 		logger.Info("Reconciling NIMService instance", "nimservice", nimService.GetName())
 		result, err := reconciler.reconcileNIMService(ctx, nimService)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


shivamerla (dismissed):
LGTM

internal/controller/platform/standalone/standalone.go
```diff
@@ -115,6 +115,10 @@ func (s *Standalone) Sync(ctx context.Context, r shared.Reconciler, resource cli
 		logger.Info("Reconciling NIMService instance", "nimservice", nimService.GetName())
 		result, err := reconciler.reconcileNIMService(ctx, nimService)
 		if err != nil {
+			if !errors.IsConflict(err) {
```
shivamerla, line 118: doesn't this need to be to return here on conflict errors instead?

varunrsekar: fixed the logic there. its supposed to immediately retry on conflict

shivamerla (approved):
Looks good

PR #505, opened 2025-05-24
Update Datastore source for NIMCache

shengnuo:
`nds-v2-huggingface-cli:25.04` has switched to using `huggingface-cli` for downloading from a HF compatible endpoint. This PR updates the NIM Operator so that it's compatible with the newer image.


api/apps/v1alpha1/nimcache_types.go
```diff
@@ -110,13 +110,17 @@ type GPUSpec struct {
 
 // DataStoreSource references a model stored on NVIDIA DataStore service.
 type DataStoreSource struct {
-	// The endpoint for datastore
+	// HF-compatible datastore endpoint
+	// +kubebuilder:validation:Pattern=`^http(s)?://.*v1/hf/?$`
```
shivamerla, line 114: we can remove the validation with `v1/hf` suffix here so that we can leverage this to pull models from HF using huggingface-cli for e.g.

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -110,13 +110,17 @@ type GPUSpec struct {
...
 	Endpoint string `json:"endpoint"`
-	// Name of either model/checkpoint or dataset to download
-	ModelName      *string `json:"modelName,omitempty"`
-	CheckpointName *string `json:"checkpointName,omitempty"`
-	DatasetName    *string `json:"datasetName,omitempty"`
-	// The name of an existing auth secret containing the AUTH_TOKEN"
+	// +kubebuilder:default="default"
+	Namespace string `json:"namespace"`
+
+	// Name of either model or dataset to download
```
shivamerla, line 119: Can we add a kubebuilder marker here to check that both model and dataset are not given at once.

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -110,13 +110,17 @@ type GPUSpec struct {
 
 // DataStoreSource references a model stored on NVIDIA DataStore service.
```
shivamerla, line 111:
```suggestion
// DataStoreSource references a model stored on NVIDIA NeMo DataStore service.
```

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -110,13 +110,17 @@ type GPUSpec struct {
...
+	// HF-compatible datastore endpoint
+	// +kubebuilder:validation:Pattern=`^http(s)?://.*$`
 	Endpoint string `json:"endpoint"`
-	// Name of either model/checkpoint or dataset to download
-	ModelName      *string `json:"modelName,omitempty"`
-	CheckpointName *string `json:"checkpointName,omitempty"`
-	DatasetName    *string `json:"datasetName,omitempty"`
-	// The name of an existing auth secret containing the AUTH_TOKEN"
+	// +kubebuilder:default="default"
+	Namespace string `json:"namespace"`
```
varunrsekar, line 117: Can you make this optional? And also add a comment as to what this field is? At first glance, we might think this is a k8s namespace.

```suggestion
	Namespace string `json:"namespace,omitempty"`
```

config/samples/nim/llm/nimcache-llm.yaml
```diff
@@ -1,7 +1,27 @@
 apiVersion: apps.nvidia.com/v1alpha1
 kind: NIMCache
 metadata:
-  name: meta-llama3-8b-instruct
+  name: meta-llama3-1b-instruct-datastore
```
varunrsekar, line 4: I'm not sure if this is the correct place for this  This probably needs to be in its own folder with a README explaining the prerequisites to run this sample.

varunrsekar (changes requested):

config/samples/nim/llm/nimcache-llm.yaml
shengnuo: This is just a sample CR to showcase how model/dataset can be pulled from an HF API compatible datastore.
How user can upload to their datastore is dependent on what they use. Docs from huggingface hub or NeMo Datastore can best guide them.

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -58,13 +58,53 @@ type NIMCacheSpec struct {
...
 type NIMSource struct {
 	// NGCSource represents models stored in NGC
 	NGC *NGCSource `json:"ngc,omitempty"`
 
-	// NGCSource represents models stored in NVIDIA DataStore service
-	DataStore *DataStoreSource `json:"dataStore,omitempty"`
+	// DataStore represents models stored in NVIDIA NeMo DataStore service
+	DataStore *NemoDataStoreSource `json:"dataStore,omitempty"`
+	// HuggingFaceHub represents models stored in HuggingFace Hub
+	HuggingFaceHub *HuggingFaceHubSource `json:"huggingFaceHub,omitempty"`
```
shivamerla, line 70:
```suggestion
	HF *HuggingFaceHubSource `json:"hf,omitempty"`
```

shivamerla (approved):
LGTM

PR #513, opened 2025-05-30
Replace the deprecated backend with model_type tags for non LLM nims

visheshtanksale:


shivamerla (dismissed):

internal/nimparser/v2/nimparser.go
```diff
@@ -80,8 +80,21 @@ func (manifest NIMManifest) MatchProfiles(modelSpec appsv1alpha1.ModelSpec, disc
...
+		// Use "model_type" if "llm_engine" is empty for non LLM models
+		if backend == "" {
+			backend = profile.Tags["model_type"]
+		}
+
+		// Fallback to deprecated "backend" tag for non LLM Models
 		if backend == "" {
-			backend = profile.Tags["backend"]
+			if deprecatedBackend, ok := profile.Tags["backend"]; ok {
+				if deprecatedBackend == "triton" {
```
shivamerla, line 92: we have to move this out of the if block as model_type can be triton.

shivamerla (approved):

PR #518, opened 2025-06-02
Allow setting PVC annotations while creation and mark various immutable fields

shivamerla:


varunrsekar (commented):
LGTM! Couple of notes:
* Please update the PR title.
* Would be good to create an issue to track the validation webhook for immutability.
* Can you add an unittest to verify the annotations are getting added?

visheshtanksale:
LGTM

PR #522, opened 2025-06-04
Change yaml package to sigs.k8s.io/yaml

shengnuo:


varunrsekar:
What's the goal of this change? None of the yaml operations we do is with k8s objects. So wasnt sure why we need to replace them. A description will be helpful.

Also, if we do need this, we should eliminate the `yaml` tags in the structs of nimparser. Since the k8s yaml pkg works with the json tags.

shivamerla:
> What's the goal of this change? None of the yaml operations we do is with k8s objects. So wasnt sure why we need to replace them. A description will be helpful.
> 
> Also, if we do need this, we should eliminate the `yaml` tags in the structs of nimparser. Since the k8s yaml pkg works with the json tags.

@varunrsekar this is in the context of the issue we have seen when marshaling PodSecurityContext for customizer config. I think the change in nimparser is fine as well as we define both yaml and json tags there, so this won't break anything but will be consistent across.

shivamerla (approved):
LGTM

PR #526, opened 2025-06-09
Adding support for building engine for buildable profiles

visheshtanksale:
Added support for NIMBuild CRD to start pod that can build a engine

**To Do**
- Add Unit test Coverage
- Add support for adding details of the new local build engine on NIMCache Status


internal/controller/nimcache_controller.go
```diff
@@ -817,6 +818,99 @@ func (r *NIMCacheReconciler) reconcileJobStatus(ctx context.Context, nimCache *a
...
+func (r *NIMCacheReconciler) reconcileEngineBuild(ctx context.Context, nimCache *appsv1alpha1.NIMCache) error {
+	logger := r.GetLogger()
+	// If the NIMCache is not in a state to build engines, return early
+	// Wait for caching to complete before building engines
+	if meta.IsStatusConditionTrue(nimCache.Status.Conditions, appsv1alpha1.NimCacheConditionJobCompleted) {
+		buildableProfiles := getBuildableProfiles(nimCache)
+
+		if buildableProfiles != nil {
+			switch {
+			case len(buildableProfiles) > 1:
```
shivamerla, line 830: This will become limiting for us as all LLM NIMs support multiple buildable profiles. Lets build using the new API as we discussed.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,657 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to update NIMBuild status", "NIMBuild", nimBuild.Name)
+		return ctrl.Result{}, err
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *NIMBuildReconciler) reconcileEngineBuild(ctx context.Context, nimBuild *appsv1alpha1.NIMBuild, nimCache *appsv1alpha1.NIMCache) error {
+	logger := r.GetLogger()
+	// If the NIMCache is not in a state to build engines, return early
```
shivamerla, line 305: nit, these two comments can be removed as the check is already enforced by the caller

visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,657 @@
...
+		}
+	} else {
+		foundProfile := getBuildableProfileByName(nimCache, nimBuild.Spec.ProfileName)
+		if foundProfile == nil {
+			logger.Info("No buildable profiles found", "ProfileName", nimBuild.Spec.ProfileName)
+			conditions.UpdateCondition(&nimBuild.Status.Conditions, appsv1alpha1.NimBuildConditionMultipleBuildableProfilesFound, metav1.ConditionTrue, "MultipleBuildableProfilesFound", "Multiple buildable profiles found, please select one profile to build")
+			nimBuild.Status.State = appsv1alpha1.NimBuildStatusFailed
+			return nil
+		} else {
+			logger.Info("Single buildable profile found", "Profile", foundProfile.Name)
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,657 @@
...
+					Name:  "NIM_SERVER_PORT",
+					Value: "8000",
+				},
+				{
+					Name:  "NIM_HTTP_API_PORT",
+					Value: "8000",
+				},
+				{
+					Name:  "NIM_CUSTOM_MODEL_NAME",
+					Value: nimCache.Name + "-" + nimProfile.Name,
```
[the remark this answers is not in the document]
visheshtanksale: Use nimProfile.Name would be confusing. Also the profile selection logic on NIMs might also fail because of duplicate name.
We can add customName as input on CRD. It can be empty if its empty then will use something auto generated

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,657 @@
...
+	return ctrl.Result{}, nil
+}
+
+func (r *NIMBuildReconciler) reconcileEngineBuild(ctx context.Context, nimBuild *appsv1alpha1.NIMBuild, nimCache *appsv1alpha1.NIMCache) error {
+	logger := r.GetLogger()
+	// If the NIMCache is not in a state to build engines, return early
+	// Wait for caching to complete before building engines
+	buildableProfile := appsv1alpha1.NIMProfile{}
+	if nimBuild.Spec.ProfileName == "" {
+		buildableProfiles := getBuildableProfiles(nimCache)
```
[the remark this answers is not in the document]
visheshtanksale: Yes we can add this when implementing Universal. We identify that the NIMCache is for Universal NIM and bypass all this logic

internal/controller/nimbuild_controller.go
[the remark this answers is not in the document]
shivamerla: cool, sounds good. this can be a follow up.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,956 @@
...
+	// To Do: Explore changing the profile list to a map for faster lookups
+	for _, profileName := range manifest.GetProfilesList() {
+		for _, selectedProfile := range nimCache.Status.Profiles {
+			if selectedProfile.Name == profileName {
+				break
+			}
+		}
+		logger.Info("Adding profile to NIMCache status", "profileName", profileName)
+		// If the profile is not found, add it to the status
+		nimCache.Status.Profiles = append(nimCache.Status.Profiles, appsv1alpha1.NIMProfile{
```
shivamerla, line 762: Looks like we end up adding duplicate profiles here. If at the end of build process, a new profile is created, then we should relate that to the original profile ID selected (or user provided). Only that new profile ID should be added in the NIMCache status with the build tags.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,956 @@
...
+	}
+	return ctrl.Result{}, nil
+}
+
+func (r *NIMBuildReconciler) reconcileEngineBuild(ctx context.Context, nimBuild *appsv1alpha1.NIMBuild, nimCache *appsv1alpha1.NIMCache) error {
+	logger := r.GetLogger()
+	buildableProfile := appsv1alpha1.NIMProfile{}
+	if nimBuild.Spec.ProfileName == "" {
+		buildableProfiles := getBuildableProfiles(nimCache)
+		if buildableProfiles != nil {
```
varunrsekar, line 330:
```suggestion
		if len(buildableProfiles) > 0 {
```

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,956 @@
...
+			return nil
+		} else {
+			logger.Info("Selected buildable profile found", "Profile", foundProfile.Name)
+			conditions.UpdateCondition(&nimBuild.Status.Conditions, appsv1alpha1.NimBuildConditionSingleBuildableProfilesFound, metav1.ConditionTrue, "BuildableProfileFound", "Single buildable profile found")
+			buildableProfile = *foundProfile
+		}
+	}
+
+	pod := &corev1.Pod{}
+	podName := types.NamespacedName{Name: nimBuild.Name + "-engine-build-pod", Namespace: nimBuild.GetNamespace()}
```
varunrsekar, line 368: Can this be a job instead? We'll have clearer understanding about the state of the build from a job and the lifecycle is easier to manage

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,116 @@
...
+	NIMCacheRef string `json:"nimCacheRef"`
+	// Required: Profile name for this engine build
+	ProfileName string `json:"profileName,omitempty"`
+	// Required: Number of GPUs to build the engine with
+	Parallelism int `json:"parallelism,omitempty"`
+	// Optional: Any user-defined tags for tracking builds
+	Tags map[string]string `json:"tags,omitempty"`
+	// Optional: Additional build params
+	BuildParams []string `json:"additionalBuildParams,omitempty"`
+}
```
varunrsekar, line 38: Can you add `NodeSelector`, `Tolerations`, `Resources` and `DRAResources` here? My understanding is that the build job would need extra memory and explicit GPUs to determine where to run.

A NIMCache's values for these might not be suitable for running the NIMBuild since they'd want to target specific GPUs.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,956 @@
...
+					Name:      "nim-cache-volume",
+					MountPath: "/model-store",
+					SubPath:   nimCache.Spec.Storage.PVC.SubPath,
+				},
+			},
+			Resources: corev1.ResourceRequirements{
+				Limits: map[corev1.ResourceName]apiResource.Quantity{
+					"cpu":            nimCache.Spec.Resources.CPU,
+					"memory":         nimCache.Spec.Resources.Memory,
+					"nvidia.com/gpu": gpuQuantity,
```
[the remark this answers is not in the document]
varunrsekar: Yeah we'd need RC support and not just rely on tp/pp detection. We bypass it the same way for nimservice.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,956 @@
...
+				},
+				Requests: map[corev1.ResourceName]apiResource.Quantity{
+					"cpu":            nimCache.Spec.Resources.CPU,
+					"memory":         nimCache.Spec.Resources.Memory,
```
varunrsekar, line 584: nimcache resources might not be sufficient as we'd need a lot more cpu/memory for the engine build.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,956 @@
...
+				if oldNIMBuild, ok := e.ObjectOld.(*appsv1alpha1.NIMBuild); ok {
+					newNIMBuild, ok := e.ObjectNew.(*appsv1alpha1.NIMBuild)
+					if ok {
+						// Handle case where object is marked for deletion
+						if !newNIMBuild.ObjectMeta.DeletionTimestamp.IsZero() {
+							return true
+						}
+
+						// Handle only spec updates
+						return !reflect.DeepEqual(oldNIMBuild.Spec, newNIMBuild.Spec)
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
visheshtanksale: The build pod doesnt run to completion. When its ready we assume that the build is complete. There is no lifecycle management difference in running it as a pod or job. So decided to run it as a pod for simplicity.

api/apps/v1alpha1/nimbuild_types.go
visheshtanksale: Updated. Will address support for DRA as another PR

internal/controller/nimbuild_controller.go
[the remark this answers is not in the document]
visheshtanksale: Will address this in a separate PR

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,956 @@
...
+	parser := nimparserutils.GetNIMParser([]byte(output))
+	// Parse the file
+	manifest, err := parser.ParseModelManifestFromRawOutput([]byte(output))
+	if err != nil {
+		logger.Error(err, "Failed to parse model manifest from the pod")
+		return err
+	}
+	logger.V(2).Info("local manifest file", "nimbuild", nimBuild.Name, "manifest", manifest)
+
+	// Create a ConfigMap with the model manifest file for re-use
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,127 @@
...
+// NIMBuildSpec to build optimized trtllm engines with given model config and weights.
+type NIMBuildSpec struct {
+	// Required: Reference to the model weights from NIMCache
+	NIMCacheRef string `json:"nimCacheRef"`
+	// Profile name for this engine build
+	Profile string `json:"profile,omitempty"`
+	// Any user-defined tags for tracking builds
+	Tags map[string]string `json:"tags,omitempty"`
+	// Additional build params
+	BuildParams []string `json:"additionalBuildParams,omitempty"`
```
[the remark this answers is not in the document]
visheshtanksale: Removed this.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1038 @@
...
+						}
+
+						// Allow only status updates
+						// Reject updates that change anything other than status
+						// Spec is immutable
+						if !reflect.DeepEqual(oldNIMBuild.Spec, newNIMBuild.Spec) {
+							return false
+						}
+						// Allow if only status changed
+						return !reflect.DeepEqual(oldNIMBuild.Status, newNIMBuild.Status)
```
varunrsekar, line 266: this would result in way too many reconciliations since the controller will update status and end up triggering this.

Spec immutability should be enforced at the API definition. And here we shouldnt reconcily any updates (or perhaps just the finalizer changes).

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,139 @@
...
+// +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Status",type=string,JSONPath=`.status.state`,priority=0
+// +kubebuilder:printcolumn:name="Age",type="date",format="date-time",JSONPath=".metadata.creationTimestamp",priority=0
+
+// NIMBuild is the Schema for the nimcaches API.
+type NIMBuild struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   NIMBuildSpec   `json:"spec,omitempty"`
```
varunrsekar, line 62:
```suggestion
       // +kubebuilder:validation:XValidation:rule="self == oldSelf", message="spec is immutable"
	Spec   NIMBuildSpec   `json:"spec,omitempty"`
```
Immutability should be handled here.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1038 @@
...
+
+	pod := &corev1.Pod{}
+	podName := types.NamespacedName{Name: nimBuild.Name + "-engine-build-pod", Namespace: nimBuild.GetNamespace()}
+	err := r.Get(ctx, podName, pod)
+	if err != nil && client.IgnoreNotFound(err) != nil {
+		return err
+	}
+	// If pod does not exist and caching is not complete, create a new one
+	if err != nil && !meta.IsStatusConditionTrue(nimBuild.Status.Conditions, appsv1alpha1.NimBuildConditionEngineBuildPodCreated) {
+		pod, err = r.constructEngineBuildPod(nimBuild, nimCache, r.orchestratorType, buildableProfile)
```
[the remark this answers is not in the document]
visheshtanksale: What is the advantage of using render package over this?

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1029 @@
...
+		conditions.UpdateCondition(&nimBuild.Status.Conditions, appsv1alpha1.NimBuildConditionWaitForNimCache, metav1.ConditionTrue, "NIMCache not ready", "Waiting for NIMCache to be ready before building engines")
+		nimBuild.Status.State = appsv1alpha1.NimBuildStatusPending
+		return ctrl.Result{}, nil
+	}
+
+	if !meta.IsStatusConditionTrue(nimBuild.Status.Conditions, appsv1alpha1.NimBuildConditionEngineBuildPodCompleted) {
+		err := r.reconcileEngineBuild(ctx, nimBuild, nimCache)
+		if err != nil {
+			logger.Error(err, "reconciliation of nimbuild failed")
+			return ctrl.Result{}, err
```
varunrsekar, line 358: When returning error here, are there any intermediate conditions that would need to be updated? In the current design, the conditions are updated only when this codepath isnt hit. Eg: `NimBuildConditionSingleBuildableProfilesFound`

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1029 @@
...
+				Model:   manifest.GetProfileModel(builtProfileName),
+				Config:  manifest.GetProfileTags(builtProfileName),
+				Release: manifest.GetProfileRelease(builtProfileName),
+			})
+		}
+	}
+
+	// Update the NIMCache status with the new profiles
+	obj := &appsv1alpha1.NIMCache{}
+	errGet := r.Get(ctx, types.NamespacedName{Name: nimCache.Name, Namespace: nimCache.GetNamespace()}, obj)
```
varunrsekar, line 839: why do we need to do another GET here?

visheshtanksale: Trying to avoid updating stale copy of the Obj.  Maybe I should add retry on conflict here https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict

internal/controller/nimbuild_controller.go
visheshtanksale: There are conditions added around the engine build pod getting created and its completion. They might get added on NIMBuild depending on where the error occurs. Also, there is a condition `NimBuildConditionReconcileFailed  ` added stating failure of reconciliation when error is returned from here

internal/controller/nimbuild_controller.go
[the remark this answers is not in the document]
visheshtanksale: Updated this on the CRD

internal/controller/nimbuild_controller.go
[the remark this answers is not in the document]
visheshtanksale: NIMCache controller does not use rendering. This was initially designed as part of NIMCache so I didnt use rendering.

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,140 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// NIMBuildSpec to build optimized trtllm engines with given model config and weights.
+type NIMBuildSpec struct {
+	// Required: Reference to the model weights from NIMCache
+	NIMCacheName string `json:"nimCacheName"`
```
shivamerla, line 30: We can make it as a NIMCacheRef to encapsulate both nimCache name and profile name.

shivamerla: We can add env as input here

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,140 @@
...
+	// Profile is the name of the profile to be used for building the engine.
+	Profile string `json:"profile,omitempty"`
+	// EngineName is the name given to the engine being built.
+	EngineName string `json:"engineName,omitempty"`
+	// Resources is the resource requirements for the NIMBuild pod.
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
shivamerla, line 36: Change this to use a custom resource type

internal/controller/nimbuild_controller.go
varunrsekar: Thinking more on this, perhaps we should return another variable `needsUpdate` in the return list of `reconcileNIMBuild` and then use that to do a single `updateNIMBuildStatus` call in `Reconcile`.

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,163 @@
...
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+	// NodeSelector is the node selector labels to schedule the caching job.
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	Env          []corev1.EnvVar   `json:"env,omitempty"`
+	Labels       map[string]string `json:"labels,omitempty"`
+	Annotations  map[string]string `json:"annotations,omitempty"`
+	ModelBuilder string            `json:"ModelBuilder"`
+	// PullSecret to pull the model builder image
+	PullSecret string `json:"pullSecret,omitempty"`
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,163 @@
...
+
+// NIMBuildStatus defines the observed state of NIMBuild.
+type NIMBuildStatus struct {
+	State         string             `json:"state,omitempty"`
+	TargetProfile NIMProfile         `json:"targetProfile,omitempty"`
+	BuiltProfile  NIMProfile         `json:"builtProfile,omitempty"`
+	Conditions    []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+}
+
+type ResourceRequirements struct {
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1033 @@
...
+
+func (r *NIMBuildReconciler) mapNIMCacheToNIMBuild(ctx context.Context, obj client.Object) []ctrl.Request {
+	nimCache, ok := obj.(*appsv1alpha1.NIMCache)
+	if !ok {
+		return []ctrl.Request{}
+	}
+
+	// Get all NIMBuilds that reference this NIMCache
+	var nimBuilds appsv1alpha1.NIMBuildList
+	if err := r.List(ctx, &nimBuilds, client.MatchingFields{"spec.nimCacheRef": nimCache.GetName()}, client.InNamespace(nimCache.GetNamespace())); err != nil {
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1033 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NIMBuildReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.recorder = mgr.GetEventRecorderFor("nimbuild-controller")
+
+	err := mgr.GetFieldIndexer().IndexField(
+		context.Background(),
+		&appsv1alpha1.NIMBuild{},
+		"spec.nimCacheName",
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

visheshtanksale:
> Follow-up: How do we account for a NIMService referencing a NIMCache for a buildable profile that is currently being built via a NIMBuild CR? In NIMService, we only check if the NIMCache is Ready before starting, but in this case we'd likely need more fine-grained checks on the profile status. And thinking crudely, we'd probably need the nimcache status to reflect the state of new profiles being built.

We cannot block NIMService from using NIMCache if there is a NIMBuild action pending. NIMCache is ready to use when its downloads the profile. There can be profile in the cache which user wants to run without building them. NIMBuild is an optional action being performed on the cache. Adding details of currently running NIMBuilds is mostly redundant because the details about relation between NIMCache and NIMBuild can also be obtained by querying NIMBuild. We add the details of successfully builds because those profile are part of the cache.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1033 @@
...
+}
+
+func (r *NIMBuildReconciler) cleanupNIMBuild(ctx context.Context, nimBuild *appsv1alpha1.NIMBuild) error {
+	var errList []error
+	logger := r.GetLogger()
+
+	// All owned objects are garbage collected
+
+	// Fetch the pod
+	podName := types.NamespacedName{Name: nimBuild.Name + "-engine-build-pod", Namespace: nimBuild.Namespace}
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,170 @@
...
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// NIMBuildSpec to build optimized trtllm engines with given model config and weights.
+type NIMBuildSpec struct {
+	// NIMCache is Reference to the model weights from NIMCache
+	NIMCache NIMCacheReference `json:"nimCache"`
+	// ModelName is the name given to the locally built engine.
+	ModelName string `json:"engineName,omitempty"`
```
varunrsekar, line 34:
```suggestion
	ModelName string `json:"modelName,omitempty"`
```

api/apps/v1alpha1/nimbuild_types.go
```diff
@@ -0,0 +1,170 @@
...
+	Labels      map[string]string `json:"labels,omitempty"`
+	Annotations map[string]string `json:"annotations,omitempty"`
+	Image       Image             `json:"image"`
+}
+
+// NIMBuildStatus defines the observed state of NIMBuild.
+type NIMBuildStatus struct {
+	State         string             `json:"state,omitempty"`
+	TargetProfile NIMProfile         `json:"targetProfile,omitempty"`
+	BuiltProfile  NIMProfile         `json:"builtProfile,omitempty"`
```
varunrsekar, line 52: `TargetProfile` is confusing. Perhaps call it `BaseProfile`/`InputProfile` and `BuiltProfile` -> `OutputProfile`?

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1035 @@
...
+	previousStatusState := nimBuild.Status.State
+
+	defer func() {
+		if err != nil {
+			r.GetEventRecorder().Eventf(nimBuild, corev1.EventTypeWarning, "ReconcileFailed",
+				"NIMBuild %s reconcile failed, msg: %s", nimBuild.Name, err.Error())
+		} else if previousStatusState != nimBuild.Status.State {
+			r.GetEventRecorder().Eventf(nimBuild, corev1.EventTypeNormal, nimBuild.Status.State,
+				"NIMBuild %s reconcile success, new state: %s", nimBuild.Name, nimBuild.Status.State)
+		}
```
varunrsekar, line 130: I'm not sure we need to emit this event - this might get too chatty.

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1035 @@
...
+				return ctrl.Result{}, err
+			}
+
+			// Remove finalizer to allow for deletion
+			controllerutil.RemoveFinalizer(nimBuild, NIMBuildFinalizer)
+			if err := r.Update(ctx, nimBuild); err != nil {
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{}, nil
+		}
```
varunrsekar, line 155:
```suggestion
                }
		return ctrl.Result{}, nil
```
You're missing the return to prevent reconciling deleting obj

api/apps/v1alpha1/nimbuild_types.go
visheshtanksale: Fixed

api/apps/v1alpha1/nimbuild_types.go
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
```diff
@@ -0,0 +1,1033 @@
...
+	}
+
+	// Handle nim-cache reconciliation
+	result, err = r.reconcileNIMBuild(ctx, nimBuild)
+	if err != nil {
+		logger.Error(err, "error reconciling NIMBuild", "name", nimBuild.Name)
+		conditions.UpdateCondition(&nimBuild.Status.Conditions, appsv1alpha1.NimBuildConditionReconcileFailed, metav1.ConditionTrue, "ReconcileFailed", err.Error())
+		nimBuild.Status.State = appsv1alpha1.NimBuildStatusFailed
+
+		err := r.updateNIMBuildStatus(ctx, nimBuild)
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

internal/controller/nimbuild_controller.go
visheshtanksale: I prefer updating the state next to the condition. I makes it more cleaner. Fixed this

shivamerla (approved):

internal/controller/nimbuild_controller.go
visheshtanksale: Addressing this as separate issue across all of NIM Operator code. Created https://jirasw.nvidia.com/browse/CNT-5981

PR #528, opened 2025-06-10
Support basic multi-node NIMs with LeaderWorkerSet

shengnuo:
This PR adds an early support of multi-node NIMs using LeaderWorkerSet.

In NIMService.spec, if `.multiNode` is specified and not nil, a LeaderWorkerSet will be created for the NIMService.


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -85,6 +91,26 @@ type NIMServiceSpec struct {
...
 	Proxy            *ProxySpec `json:"proxy,omitempty"`
+	// MultiNode config for NIMService using LeaderWorkerSet
+	MultiNode *NimServiceMultiNodeConfig `json:"multiNode,omitempty"`
+}
+
+// NimServiceMultiNodeConfig defines the configuration for multi-node NIMService.
+// TODO disable Scale when this is enabled
+type NimServiceMultiNodeConfig struct {
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:default:=2
```
shivamerla, line 102: I thought we will not limit it to minimum 2 workers to allow testing on single node correct?

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -85,6 +91,26 @@ type NIMServiceSpec struct {
...
+	// Workers specifies how many worker pods per multi-node replica to launch.
+	// +kubebuilder:validation:Minimum=2
+	Workers int `json:"workers,omitempty"`
+
+	// +kubebuilder:default:=1
+	// GPUPerNode specifies the number of GPUs for each pod. In most cases, this should match `resources.limits.nvidia.com/gpu`.
+	GPUPerNode int `json:"gpusPerNode,omitempty"`
+
+	// +kubebuilder:default:=300
```
shivamerla, line 111: I think these should be moved into a separate  MPIConfig or MPIParams struct.

internal/controller/nimservice_controller.go
```diff
@@ -238,6 +240,7 @@ func (r *NIMServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&appsv1alpha1.NIMService{}).
 		Owns(&appsv1.Deployment{}).
+		Owns(&lwsv1.LeaderWorkerSet{}).
```
varunrsekar, line 243: This needs to be only added if LWS CRDs are installed in the cluster. Please use the `CRDExists` util to add this conditionally.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -364,7 +515,7 @@ func (r *NIMServiceReconciler) getNIMModelName(ctx context.Context, nimServiceEn
 	modelsList, err := nimmodels.ListModelsV1(ctx, nimServiceEndpoint)
 	if err != nil {
 		logger.Error(err, "Failed to list models", "endpoint", nimServiceEndpoint)
-		// Check if it's an HTTP error with a status code
+		// Check if iterr's an HTTP error with a status code
```
varunrsekar, line 518: nit: typo

internal/controller/platform/standalone/nimservice.go
```diff
@@ -483,21 +634,9 @@ func (r *NIMServiceReconciler) getNIMModelEndpoints(ctx context.Context, nimServ
...
-	if err != nil && !errors.IsNotFound(err) {
-		logger.Error(err, fmt.Sprintf("Error is not NotFound for %s: %v", obj.GetObjectKind(), err))
-		return err
-	}
-	// Don't do anything if CR is unchanged.
-	if err == nil && !utils.IsParentSpecChanged(obj, utils.DeepHashObject(nimService.Spec)) {
-		return nil
-	}
-
 	resource, err := renderFunc()
```
varunrsekar, line 637: Please move the rendering below the CR spec comparison

internal/controller/platform/standalone/nimservice.go
```diff
@@ -543,6 +699,24 @@ func (r *NIMServiceReconciler) renderAndSyncResource(ctx context.Context, nimSer
 	return nil
 }
 
+func (r *NIMServiceReconciler) isLeaderWorkerSetReady(ctx context.Context, nimService *appsv1alpha1.NIMService) (string, bool, error) {
```
varunrsekar, line 702: Its strange for this function to return a string in its current form. Do you expect other states than just Ready/NotReady to be possible for this? Otherwise, lets just start with `(bool, error)` and then grow this later.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -698,6 +872,47 @@ func (r *NIMServiceReconciler) getTensorParallelismByProfile(ctx context.Context
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse tensorParallelism: %w", err)
+		}
+	}
+	resources.Requests[gpuResourceName] = gpuQuality
+	resources.Limits[gpuResourceName] = gpuQuality
+	return &resources, nil
+}
+
 // assignGPUResources automatically assigns GPU resources to the NIMService based on the provided profile,
```
varunrsekar, line 916: If this function is unused, please remove it.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -698,6 +872,47 @@ func (r *NIMServiceReconciler) getTensorParallelismByProfile(ctx context.Context
 	return tensorParallelism, nil
 }
 
+func (r *NIMServiceReconciler) getGPUResources(ctx context.Context, nimService *appsv1alpha1.NIMService, profile *appsv1alpha1.NIMProfile) (*corev1.ResourceRequirements, error) {
```
varunrsekar, line 875: Please modify the `assignGPUResources` unittests to be for this function.

varunrsekar (changes requested):

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -80,11 +96,43 @@ type NIMServiceSpec struct {
...
+	MultiNode        *NimServiceMultiNodeConfig `json:"multiNode,omitempty"`
+}
+
+// NimServiceMultiNodeConfig defines the configuration for multi-node NIMService.
+// TODO disable Scale when this is enabled
+type NimServiceMultiNodeConfig struct {
+	// +kubebuilder:validation:Enum=lws
+	// +kubebuilder:default:=lws
+	// BackendType specifies the backend type for the multi-node NIMService. Currently only LWS is supported.
+	BackendType string `json:"backendType,omitempty"`
```
varunrsekar, line 113: nit: Please define a separate type alias for this with pre-defined consts.

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -80,11 +96,43 @@ type NIMServiceSpec struct {
...
+	GPUPerWorker int `json:"gpusPerWorker,omitempty"`
+
+	// MPI config for NIMService using LeaderWorkerSet
+	MPI *MultiNodeMPIConfig `json:"mpi,omitempty"`
+}
+
+type MultiNodeMPIConfig struct {
+	// CustomMPIStartScript specifies the custom MPI start script to use.
+	// A default start script will be used if this is not specified.
+	CustomMPIStartScript string `json:"customMPIStartScript,omitempty"`
```
varunrsekar, line 131: I've reservations about allowing freeform code here. IMO, this should not be an input. If at all there is a need for a custom script, it should be in a configmap and that needs to be input here (But I'd rather not support this)

internal/controller/platform/standalone/nimservice.go
shengnuo: no longer possible.
the previous implementation assumed that all supporting objects are named exactly like the parent NIMService. This assumption no longer holds for LWS, so we must render, so we know what to compare with.

api/apps/v1alpha1/nimservice_types.go
shengnuo: ya it's better to provide customizable MPI script as a feature if customer requests it. Removed

internal/k8sutil/k8sutil.go
```diff
@@ -288,3 +291,30 @@ func GetRawYAMLFromConfigMap(ctx context.Context, k8sClient client.Client, names
 
 	return raw, nil
 }
+
+func CRDExists(crdName string) (bool, error) {
```
varunrsekar, line 295: Can you sync to ToT? This method already exists

config/samples/nim/llm/nimservice.yaml
```diff
@@ -22,3 +22,8 @@ spec:
     service:
       type: ClusterIP
       port: 8000
+  multiNode:
```
shivamerla, line 25: we should add a separate sample for this than the default sample.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -176,6 +176,7 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
 
 	// Sync serviceaccount
 	err = r.renderAndSyncResource(ctx, nimService, &renderer, &corev1.ServiceAccount{}, func() (client.Object, error) {
+		logger.Info("Rendering serviceaccount", "nimservice", nimService.Name)
```
shivamerla, line 179: This will be too verbose if we log for every resource. We need to move this to higher log levels.

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -86,11 +108,38 @@ type NIMServiceSpec struct {
...
+type NimServiceMultiNodeConfig struct {
+	// +kubebuilder:validation:Enum=lws
+	// +kubebuilder:default:=lws
+	// BackendType specifies the backend type for the multi-node NIMService. Currently only LWS is supported.
+	BackendType NIMBackendType `json:"backendType,omitempty"`
+
+	// +kubebuilder:default:=1
+	// Workers specifies how many worker pods per multi-node replica to launch.
+	// +kubebuilder:validation:Minimum=1
+	Workers int `json:"workers,omitempty"`
```
varunrsekar, line 129: Can we call this `Size` similar to the LWS CRD? `Workers` doesnt immediately tell us if it includes the leader or not.

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -155,6 +204,11 @@ type NIMServiceStorage struct {
 	ReadOnly *bool `json:"readOnly,omitempty"`
 }
 
+// GetLWSName returns the name to be used for the LeaderWorkerSet based on the custom spec
+func (n *NIMService) GetLWSName() string {
+	return fmt.Sprintf("%s-lws", n.GetName())
```
varunrsekar, line 209: is the `-lws` suffix needed? Do you expect any collisions if just use `GetName()`?

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -86,11 +108,38 @@ type NIMServiceSpec struct {
...
+	BackendType NIMBackendType `json:"backendType,omitempty"`
+
+	// +kubebuilder:default:=1
+	// Workers specifies how many worker pods per multi-node replica to launch.
+	// +kubebuilder:validation:Minimum=1
+	Workers int `json:"workers,omitempty"`
+
+	// +kubebuilder:default:=1
+	// GPUPerWorker specifies the number of GPUs for each worker. In most cases, this should match `resources.limits.nvidia.com/gpu`.
+	GPUPerWorker int `json:"gpusPerWorker,omitempty"`
```
varunrsekar, line 133: Do we need this? if we expect `spec.resources.requests[nvidia.com/gpu]` to be the gpus per worker, it might be simpler. 
For auto-detection, we'd use `tp` size anyway (or default to 1).

We can compute the `GPUS_PER_NODE` env at reconcile-time. DRA would need more thought on how to match GPUPerWorker with the claims but I dont see why we need this input for DP.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -831,34 +998,30 @@ func (r *NIMServiceReconciler) assignGPUResources(ctx context.Context, nimServic
...
 	}
 
-	// Assign GPU resources based on tensorParallelism, or default to 1 GPU if tensorParallelism is not available
-	gpuQuantity := apiResource.MustParse("1") // Default to 1 GPU
-
+	gpuQuality := apiResource.MustParse("1")
+	if nimService.Spec.MultiNode != nil && nimService.Spec.MultiNode.GPUPerWorker > 0 {
+		gpuQuality = apiResource.MustParse(fmt.Sprintf("%d", nimService.Spec.MultiNode.GPUPerWorker))
+	}
 	if tensorParallelism != "" {
```
varunrsekar, line 1018: In line with my api comment, its weird that we use the GPUPerWorker input but override it by `tp` size immediately after. Lets drop it

varunrsekar (commented):
Overall looks good to me. Just some more api comments...

api/apps/v1alpha1/nimservice_types.go
shengnuo: we create a Service named exactly like the Nimservice, this would conflict with the headless service created by LWS, if the LWS is also named exactly like the NIMService.

The headless service is needed to provide a domain name for each pod, which is used by MPI for pod to pod communication

api/apps/v1alpha1/nimservice_types.go
shengnuo: Ya I agree this can be omitted if we only use DP.
In the case of DRA, this might not be too straightforward, especially in GB200 system, where we claim compute domains rather than individual GPUs.

So I am keeping this field around for the time being, until we can find a way to auto compute this value in the follow up.

internal/controller/platform/standalone/nimservice.go
shengnuo: The total GPU needed is `tp*pp`, and GPUs per worker is really just `tp*pp/workers`.

In single-node monolithic NIMs, since pp=workers=1, we can assign the GPU numbers equal to tp.
But for multinode, we would have to treat it differently.

For now, I am setting the number of GPU resources equal to `.spec.multiNode.GPUSPerPod` per pod.
If that field were to deprecated in future, we can calculate using `tp*pp/workers`.

api/apps/v1alpha1/nimservice_types.go
shengnuo: I have a feeling this field might be needed for DRA.

Will reconsider this field in a follow up.

api/apps/v1alpha1/nimservice_types.go
shengnuo: Not sure how will this work when we enable auto scaling.
Lets circle back in a follow up

shivamerla (dismissed):

varunrsekar (dismissed):
Everything else lgtm. Please fix the pending comment ~before merge~ in a follow-up

shivamerla (approved):

PR #531, opened 2025-06-12
Changes to support NeMo Evaluator 25.06

shivamerla:
* New evaluation images support (BFCL, Agentic Evaluation)
* Remove the evaluation image validation markers (as some are optional across versions)
* Update Unit tests


api/apps/v1alpha1/nemo_evaluator_types.go
```diff
@@ -197,6 +193,14 @@ func (ei EvaluationImages) GetEvaluationImageEnv() []corev1.EnvVar {
 			Name:  "RAG",
 			Value: ei.Rag,
 		},
+		{
```
varunrsekar, line 196: Shouldn't we only set each of these env vars if the value was provided in EI?

internal/controller/nemocustomizer_controller.go
```diff
@@ -557,10 +567,21 @@ func (r *NemoCustomizerReconciler) addModelDownloadJobsConfig(ctx context.Contex
 		"imagePullSecrets":        pullSecrets,
 		"ngcAPISecret":            n.Spec.ModelDownloadJobs.NGCSecret.Name,
 		"ngcAPISecretKey":         n.Spec.ModelDownloadJobs.NGCSecret.Key,
+		"ngcSecretName":           n.Spec.ModelDownloadJobs.NGCSecret.Name,
+		"ngcSecretKey":            n.Spec.ModelDownloadJobs.NGCSecret.Key,
```
varunrsekar, line 571: Shouldnt these be set only if NGCSecret is provided? And be mutually exclusive with HFSecret?

internal/controller/nemocustomizer_controller.go
```diff
@@ -694,6 +719,44 @@ func (r *NemoCustomizerReconciler) addModelConfig(ctx context.Context, cfg map[s
 	return nil
 }
 
+func (r *NemoCustomizerReconciler) addCustomizationTargetConfig(ctx context.Context, cfg map[string]interface{}, n *appsv1alpha1.NemoCustomizer) error {
```
varunrsekar, line 722: nit: reorder to put changing `cfg` at the end of the params list

internal/controller/nemocustomizer_controller.go
```diff
@@ -694,6 +719,44 @@ func (r *NemoCustomizerReconciler) addModelConfig(ctx context.Context, cfg map[s
...
+	var customizationTargets map[string]interface{}
+	if err := yaml.Unmarshal([]byte(customizationTargetsRaw), &customizationTargets); err != nil {
+		return fmt.Errorf("parsing customization targets config: %w", err)
+	}
+
+	cfg["customizationTargets"] = customizationTargets
+	return nil
+}
+
+func (r *NemoCustomizerReconciler) addCustomizationConfigTemplates(ctx context.Context, cfg map[string]interface{}, n *appsv1alpha1.NemoCustomizer) error {
```
varunrsekar, line 741: nit: reorder to put changing cfg at the end of the params list

api/apps/v1alpha1/nemo_evaluator_types.go
shivamerla: all these images are mandatory for evaluator without which the eval pod is crashing. So, we will need to set all of these and later in validations we need to enforce that.

internal/controller/nemocustomizer_controller.go
shivamerla: i am adding NGC keys by default as all default configs are from NGC. There are no HF targets in the model config by default. Also, we can set both at the same time, depending on the target type, particular token is used.

api/apps/v1alpha1/nemo_evaluator_types.go
varunrsekar: Thanks for the clarification. I thought some of these images were optional.

varunrsekar (approved):
LGTM

PR #565, opened 2025-07-03
use go yaml pkg for parsing model manifests

varunrsekar:
Model manifests for certain nims (schema_version 1.0) sometimes use custom yaml tags which cannot be parsed by the `sigs.k8s.io/yaml` pkg. 

Eg:
```
...
      workspace: !workspace
        components:
       - dst: ""
          src:
            files:
            - !name README.md
            - !name checksums.blake3
            - !name config.json
...
```


internal/controller/nimcache_controller.go
```diff
@@ -1333,18 +1333,6 @@ func (r *NIMCacheReconciler) createManifestConfigMap(ctx context.Context, nimCac
 		return fmt.Errorf("failed to marshal manifest data: %w", err)
 	}
 
-	// Pretty-print the YAML content
```
shivamerla, line 1336: Did you test out that the ConfigMap created have right fields without this? I think pointer types were not getting marshalled correctly.

shivamerla: hmm, i think due to custom tags in the v1 format, the indentation was messed up, so unmarshalling into an empty interface was fixing it. If that is no longer true with the Go yaml package, then we can ignore this.

shivamerla (approved):

internal/controller/nimcache_controller.go
varunrsekar: I believe `gopkg.in/yaml` does 4-space indentations when marshalling. But `sigs.k8s.io/yaml` does minimal indentation (2-spaces by default and lists start with 0-spaces) like we expect - so we should be okay with the change in nimcache_controller.

PR #577, opened 2025-07-14
Should give precedence to user provided env var on Universal NIM Service

visheshtanksale:
If user provides Env Vars on NIMService then those env vars should have precedence.


varunrsekar:
Thanks. This LGM...

There's a couple other discrepancies as well. Can you fix those?
* GetLWSLeaderEnv: https://github.com/NVIDIA/k8s-nim-operator/blob/main/api/apps/v1alpha1/nimservice_types.go#L304
* GetLWSWorkerEnv: https://github.com/NVIDIA/k8s-nim-operator/blob/main/api/apps/v1alpha1/nimservice_types.go#L361
* https://github.com/NVIDIA/k8s-nim-operator/blob/main/internal/controller/platform/kserve/nimservice.go

shivamerla (approved):
Good catch @visheshtanksale

PR #596, opened 2025-07-29
Fix to handle CR cleanup with foreground propagation

shivamerla:
* Handle custom cleanup irrespective of finalizer is set or not
* Remove custom finalizer if present
* Avoid further reconciliation when the object is being deleted

### DIFF
--- a/internal/controller/nemo_datastore_controller.go
+++ b/internal/controller/nemo_datastore_controller.go
@@ -133,14 +133,13 @@ func (r *NemoDatastoreReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	} else {
 		// The instance is being deleted
+		// Perform platform specific cleanup of resources
+		if err := r.cleanupNemoDatastore(ctx, NemoDatastore); err != nil {
+			r.GetEventRecorder().Eventf(NemoDatastore, corev1.EventTypeNormal, "Delete",
+				"NemoDatastore %s is being deleted", NemoDatastore.Name)
+			return ctrl.Result{}, err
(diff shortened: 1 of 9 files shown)


internal/controller/nemo_datastore_controller.go
```diff
@@ -133,23 +133,23 @@ func (r *NemoDatastoreReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		}
 	} else {
 		// The instance is being deleted
+		// Perform platform specific cleanup of resources
+		if err := r.cleanupNemoDatastore(ctx, NemoDatastore); err != nil {
+			r.GetEventRecorder().Eventf(NemoDatastore, corev1.EventTypeNormal, "Delete",
+				"NemoDatastore %s is being deleted", NemoDatastore.Name)
+			return ctrl.Result{}, err
+		}
```
varunrsekar, line 141: We dont need to do this:
1. those cleanup methods are currently no-op
2. The existing behavior where we only attempt cleanup as long as our finalizer exists is already correct.

Just shortcircuiting the reconcile loop at L152 is sufficient.

shivamerla: hmm, its still good to have the cleanup method  (with both finalizers) for blocking deletion on other operands (for e.g. is any training job is in progress or download is in progress etc). These are placeholders for such cases.

shivamerla: but in this case with the foreground propagation, it will be missing and we will have to do custom cleanup on deletion.

PR #610, opened 2025-08-13
Kserve: Fix to use the cluster endpoint for updating model status in the serverless mode as well

shivamerla:


shivamerla:
@xieshenzh PTAL

PR #614, opened 2025-08-14
Misc NIMPipeline fixes

shivamerla:
* Error out if there is a conflict with the NIM services in the pipeline vs pre-existing service
* Clenup NIM service if the spec is removed from the pipeline

### DIFF
--- a/bundle/manifests/k8s-nim-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/k8s-nim-operator.clusterserviceversion.yaml
@@ -1334,6 +1334,18 @@ spec:
               - patch
               - update
               - watch
+            - apiGroups:
+              - gateway.networking.k8s.io
+              resources:
+              - httproutes
+              verbs:
+              - create
+              - delete
+              - get
+              - list
+              - patch
+              - update
+              - watch
       deployments:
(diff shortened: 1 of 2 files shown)


varunrsekar (dismissed):

internal/controller/nimpipeline_controller.go
```diff
@@ -316,19 +308,17 @@ func (r *NIMPipelineReconciler) updateStatus(ctx context.Context, nimPipeline *a
...
-			overallState = appsv1alpha1.NIMServiceStatusFailed
-			allServicesReady = false
-		case appsv1alpha1.NIMServiceStatusNotReady, appsv1alpha1.NIMServiceStatusPending:
-			// If any service is not ready or pending, set overall readiness to false
+		case appsv1alpha1.NIMServiceStatusReady:
+			// Leave the overall status as is
+		default:
+			// If any service is not ready, set overall readiness to false
 			allServicesReady = false
 		}
```
varunrsekar, line 316: Can you get rid of `overallState` references? It seems redundant after this change

shivamerla: i had missed that, we indeed need that for failed case.

PR #619, opened 2025-08-15
Some kserve validations added.

aryangorwade:
KServe validations below added. Also added comprehensive tests for these validations.

1) When inferencePlatform is "kserve" and used in serverless mode (edited) 
- autoscaling cannot be set
- ingress cannot be set
- servicemonitor cannot be set

2) multiNode cannot be enabled when kserve inferencePlatform is set


shivamerla (approved):
LGTM, thanks @aryangorwade

PR #626, opened 2025-08-19
Made cert-manager a required dependency. Removed support for custom certificates

aryangorwade:
Removed useCertManager as well as the certificates section from Helm deployment.

### DIFF
--- a/deployments/helm/k8s-nim-operator/values.yaml
+++ b/deployments/helm/k8s-nim-operator/values.yaml
@@ -54,14 +54,9 @@ operator:
       drop:
         - ALL
   admissionController:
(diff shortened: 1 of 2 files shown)


shivamerla (approved):
LGTM, thanks @aryangorwade

deployments/helm/k8s-nim-operator/values.yaml
```diff
@@ -56,12 +56,6 @@ operator:
   admissionController:
```
shivamerla, line 56: May be add a comment here as below.

```
# Enable the admission controller. 
# Note: cert-manager must be installed beforehand, as it is required to generate the TLS certificates.
```

PR #631, opened 2025-08-22
Cleanup Multi Node API to avoid redundant GPU input

visheshtanksale:


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -227,6 +223,16 @@ func (n *NIMService) GetLWSName() string {
 	return fmt.Sprintf("%s-lws", n.GetName())
 }
 
+// GetMultiNodeGPUsPerPod returns the number of GPUs per pod for the multi-node NIMService.
+func (n *NIMService) GetMultiNodeGPUsPerPod() int {
+	gpuQuantity, ok := n.Spec.Resources.Requests["nvidia.com/gpu"]
```
varunrsekar, line 228: do we need to support DRA here?

internal/shared/multinode.go
```diff
@@ -0,0 +1,49 @@
...
+
+	appsv1alpha1 "github.com/NVIDIA/k8s-nim-operator/api/apps/v1alpha1"
+)
+
+// GetMultiNodeGPUsPerPod returns the number of GPUs per pod for the multi-node NIMService.
+// If the NIMService is not a multi-node NIMService, it returns the number of GPUs per pod based on the user-provided GPU resource requests.
+// If the NIMService is a multi-node NIMService, it returns the number of GPUs per pod based on the DRA resources.
+func GetMultiNodeGPUsPerPod(ctx context.Context, client client.Client, nimService *appsv1alpha1.NIMService) (int, error) {
+
+	if nimService.Spec.MultiNode == nil {
```
shivamerla, line 32: This check doesn't seem right. If multi-node is not enabled, then we can return error from this call as we cannot deploy with resource count of 0. Also, it is better to make this call generic i.e GetGPUCountPerPod() which returns the count of full GPUs from either the resource spec or from DRA resources.

internal/controller/platform/kserve/nimservice.go
```diff
@@ -577,7 +583,7 @@ func (r *NIMServiceReconciler) addGPUResources(ctx context.Context, nimService *
 	// if deployed as multi-node, use the GPU per worker value to assign GPU resources to each worker
 	// TODO auto determine base on tp*pp/(.spec.multiNode.size)
 	if nimService.Spec.MultiNode != nil {
```
shivamerla, line 585: i think this entire multi-node check should be removed. This function is for auto assignment of GPUs, which holds true for multi-node as well when a trtllm profile is selected. We don't support DRA for auto-assignment anyway.

internal/shared/multinode.go
visheshtanksale: Fixed it. The check was supposed to be for DRA instead of Multi Node

internal/controller/platform/standalone/nimservice.go
```diff
@@ -349,6 +349,12 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
 	var conType, failedCon string
 	var renderObj client.Object
 
+	gpuCountPerPod, err := shared.GetGPUCountPerPod(ctx, r.GetClient(), nimService)
```
shengnuo, line 352: Would this function need to be invoked each time we reconcile a NIMService object?
Can we cache the GPU count per pod for DRA, so we can avoid repeatly calling this function.

An idea would be annotation to the NIMService object for GPUCountPerPod, and only run this function if the annotation is not present.

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -1087,7 +1083,7 @@ func (n *NIMService) GetDeploymentParams() *rendertypes.DeploymentParams {
 	return params
 }
 
-func (n *NIMService) GetLWSParams() *rendertypes.LeaderWorkerSetParams {
+func (n *NIMService) GetLWSParams(multiNodeGPUsPerPod int) *rendertypes.LeaderWorkerSetParams {
```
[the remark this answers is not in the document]
shengnuo: Is it possible to cache this gpuPerPod info as part of the NIMService object (either as annotation, or part of the nimservice status)?

In this way, gpuPerPod information would be part of the NIMservice object itself, without needing to pass the parameter around multiple functions.

internal/shared/resourceclaims.go
```diff
@@ -228,3 +230,138 @@ func GetDRADeviceCELExpressions(device appsv1alpha1.DRADeviceSpec) ([]string, er
...
+		var err error
+
+		switch {
+		case resource.ResourceClaimName != nil:
+			gpuCount, err = GetGPUDeviceCountForClaim(ctx, client, *resource.ResourceClaimName, namespace)
+		case resource.ResourceClaimTemplateName != nil:
+			gpuCount, err = GetGPUDeviceCountForClaimTemplate(ctx, client, *resource.ResourceClaimTemplateName, namespace)
+		case resource.ClaimCreationSpec != nil:
+			gpuCount, err = GetGPUDeviceCountForClaimCreationSpec(ctx, client, *resource.ClaimCreationSpec, namespace)
+		}
```
shengnuo, line 270: I am wondering do we have a use case for users to specify two of (resourceClaimName, resoruceClaimTemplateName, claimCreationSpec)...?

I thought users would only specify one of three?
If that's the case, we should consider adding to the validation webhook

internal/shared/multinode.go
```diff
@@ -0,0 +1,47 @@
...
+
+	appsv1alpha1 "github.com/NVIDIA/k8s-nim-operator/api/apps/v1alpha1"
+)
+
+// GetGPUCountPerPod returns the number of GPUs per pod for the NIMService.
+func GetGPUCountPerPod(ctx context.Context, client client.Client, nimService *appsv1alpha1.NIMService) (int, error) {
+
+	if nimService.Spec.DRAResources == nil {
+		if nimService.Spec.Resources == nil {
+			return 0, nil
```
varunrsekar, line 32: we should return an error when no GPUs are specified and fail the NIMService for multiNode.

internal/shared/multinode.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+		gpuQuantity, ok := nimService.Spec.Resources.Requests["nvidia.com/gpu"]
+		if !ok {
+			// return 0 if no GPU limit is specified because auto determine base on tp*pp/(.spec.multiNode.size) is a TODO
+			return 0, nil
+		}
```
varunrsekar, line 38: can you also check for resource limits here?

internal/shared/resourceclaims.go
```diff
@@ -228,3 +230,138 @@ func GetDRADeviceCELExpressions(device appsv1alpha1.DRADeviceSpec) ([]string, er
...
+	NVIDIAComIdentifier = "nvidia.com"
+)
+
+// isNVIDIAGPU checks if a device class represents an NVIDIA GPU.
+func isNVIDIAGPU(ctx context.Context, client client.Client, deviceClassName string) (bool, error) {
+	if deviceClassName == "" {
+		return false, nil
+	}
+
+	var dc resourcev1beta2.DeviceClass
```
varunrsekar, line 247: Can you start an informercache for DeviceClass objects on startup?

internal/shared/resourceclaims.go
```diff
@@ -228,3 +230,138 @@ func GetDRADeviceCELExpressions(device appsv1alpha1.DRADeviceSpec) ([]string, er
...
+	if err := client.Get(ctx, types.NamespacedName{Name: deviceClassName}, &dc); err != nil {
+		return false, fmt.Errorf("failed to get device class %s: %w", deviceClassName, err)
+	}
+
+	hint := strings.ToLower(fmt.Sprint(dc.Spec))
+	return strings.Contains(hint, NVIDIAIdentifier) || strings.Contains(hint, NVIDIAComIdentifier), nil
```
varunrsekar, line 253: nit: we should restrict to `gpu.nvidia.com` device class.

internal/shared/resourceclaims.go
varunrsekar: @shengnuo this is possible across the spec.draResources[] array. 1 entry may have a resourceClaimName while another one may have a claimCreationSpec. We would need to aggregate them

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -386,13 +393,45 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
-func validateResourcesConfiguration(resources *corev1.ResourceRequirements, fldPath *field.Path) field.ErrorList {
+func validateResourcesConfiguration(spec *appsv1alpha1.NIMServiceSpec, fldPath *field.Path) field.ErrorList {
 	errList := field.ErrorList{}
-	if resources != nil {
-		if resources.Claims != nil || len(resources.Claims) != 0 {
-			errList = append(errList, field.Forbidden(fldPath.Child("claims"), "must be empty"))
-		}
+
+	// Validate that Claims must be empty
+	if spec.Resources != nil && spec.Resources.Claims != nil && len(spec.Resources.Claims) > 0 {
```
varunrsekar, line 400: nit: if its nil, then len(...) will return 0
```suggestion
	if spec.Resources != nil && len(spec.Resources.Claims) > 0 {
```

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -386,13 +393,45 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
+}
+
+// validateGPURequirements ensures that GPU resources are properly configured for MultiNode deployments.
+func validateGPURequirements(spec *appsv1alpha1.NIMServiceSpec, fldPath *field.Path) field.ErrorList {
+	errList := field.ErrorList{}
+
+	gpuResourceName := corev1.ResourceName("nvidia.com/gpu")
+
+	// Check if GPU requests are specified
+	if spec.Resources == nil || spec.Resources.Requests == nil {
```
varunrsekar, line 419: nit: also check for resource limits

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -69,10 +69,17 @@ func validateNIMServiceSpec(spec *appsv1alpha1.NIMServiceSpec, fldPath *field.Pa
 	errList = append(errList, validateExposeConfiguration(&spec.Expose, fldPath.Child("expose").Child("ingress"))...)
 	errList = append(errList, validateMetricsConfiguration(&spec.Metrics, fldPath.Child("metrics"))...)
 	errList = append(errList, validateScaleConfiguration(&spec.Scale, fldPath.Child("scale"))...)
-	errList = append(errList, validateResourcesConfiguration(spec.Resources, fldPath.Child("resources"))...)
+	errList = append(errList, validateResourcesConfiguration(spec, fldPath.Child("resources"))...)
```
varunrsekar, line 72:
```suggestion
	errList = append(errList, validateResourcesConfiguration(spec, fldPath)...)
```
nit: its better to match the fieldPath with the input in the caller.

internal/shared/resourceclaims.go
visheshtanksale: Its possible for users to specify all three resource specification types

internal/controller/platform/standalone/nimservice.go
visheshtanksale: Fixed to use annotation

internal/controller/platform/standalone/nimservice.go
```diff
@@ -1004,7 +1010,7 @@ func (r *NIMServiceReconciler) addGPUResources(ctx context.Context, nimService *
 	// if deployed as multi-node, use the GPU per worker value to assign GPU resources to each worker
 	// TODO auto determine base on tp*pp/(.spec.multiNode.size)
 	if nimService.Spec.MultiNode != nil {
-		gpuQuantity, err = apiResource.ParseQuantity(fmt.Sprintf("%d", nimService.Spec.MultiNode.GPUSPerPod))
+		gpuQuantity, err = apiResource.ParseQuantity(fmt.Sprintf("%d", multiNodeGPUsPerPod))
```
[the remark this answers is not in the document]
visheshtanksale: Removed this

internal/shared/multinode.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appsv1alpha1 "github.com/NVIDIA/k8s-nim-operator/api/apps/v1alpha1"
+)
+
+// GetGPUCountPerPod returns the number of GPUs per pod for the NIMService.
+func GetGPUCountPerPod(ctx context.Context, client client.Client, nimService *appsv1alpha1.NIMService) (int, error) {
+
+	if nimService.Spec.DRAResources == nil {
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

internal/shared/multinode.go
visheshtanksale: Fixed

internal/shared/resourceclaims.go
visheshtanksale: Done

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
visheshtanksale: Fixed

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
visheshtanksale: Fixed

internal/controller/nimservice_controller.go
```diff
@@ -133,6 +134,18 @@ func (r *NIMServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	logger.Info("Reconciling", "NIMService", nimService.Name)
 
+	if nimService.Spec.MultiNode != nil && nimService.Annotations != nil {
```
shivamerla, line 137: annotations could be nil initially, we would need to initialize the map for the first time.

shivamerla: also, this should be done in the platform specific reconciler where we render the LWS resource. That way we can consider the case of auto GPU assignment too based on the tp size in the optimized profile, otherwise there would be mismatch and we end up setting this annotation to `0`.

internal/shared/resourceclaims.go
```diff
@@ -228,3 +230,137 @@ func GetDRADeviceCELExpressions(device appsv1alpha1.DRADeviceSpec) ([]string, er
...
+func getGPUCountFromDeviceRequests(ctx context.Context, client client.Client, requests []resourcev1beta2.DeviceRequest) (int, error) {
+
+	for _, req := range requests {
+		if req.Exactly.DeviceClassName == "" {
+			continue
+		}
+
+		isGPU, err := isNVIDIAGPU(ctx, client, req.Exactly.DeviceClassName)
+		if err != nil {
+			// This allows partial success scenarios
```
shivamerla, line 334: we would need to fail in all cases here as GPU count might be incorrect, if there is an error getting the device-class for e.g.

internal/controller/platform/kserve/nimservice.go
```diff
@@ -574,29 +574,19 @@ func (r *NIMServiceReconciler) addGPUResources(ctx context.Context, nimService *
...
-		}
-	} else {
-		// If no user-provided GPU resource is found, proceed with auto-assignment
-		// Get tensorParallelism from the profile
-		tensorParallelism, err := utils.GetTensorParallelismByProfileTags(profile.Config)
+	// If no user-provided GPU resource is found, proceed with auto-assignment
+	// Get tensorParallelism from the profile
+	tensorParallelism, err := utils.GetTensorParallelismByProfileTags(profile.Config)
+	if err != nil {
+		logger.Error(err, "Failed to retrieve tensorParallelism")
```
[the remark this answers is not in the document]
visheshtanksale: Fixed

internal/shared/resourceclaims.go
visheshtanksale: Added informer cache

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -386,13 +393,59 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
+}
+
+// validateGPURequirements ensures that GPU resources are properly configured for MultiNode deployments.
+func validateGPURequirements(spec *appsv1alpha1.NIMServiceSpec, fldPath *field.Path) field.ErrorList {
+	errList := field.ErrorList{}
+
+	gpuResourceName := corev1.ResourceName("nvidia.com/gpu")
+
+	// Check if GPU requests or limits are specified
+	if spec.Resources == nil {
```
shivamerla, line 419: This should also consider `spec.DRAResources`

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -385,14 +385,64 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
+
+	gpuResourceName := corev1.ResourceName("nvidia.com/gpu")
+
+	// Check if GPU requests or limits are specified or DRA resources are specified
+	if spec.Resources == nil && len(spec.DRAResources) == 0 {
+		errList = append(errList, field.Required(fldPath.Child("resources"), "GPU resources must be specified for MultiNode deployments"))
+		return errList
+	}
+
+	if spec.Resources != nil {
```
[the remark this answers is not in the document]
visheshtanksale: This check in primary on spec.resouces and we dont have any checks under spec.draResources because for that we need to make an API request. Since this includes only checks under spec.resources I feel it better to have the implementation here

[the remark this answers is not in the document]
visheshtanksale: Fixed

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -158,6 +153,13 @@ type MultiNodeMPIConfig struct {
 	MPIStartTimeout int `json:"mpiStartTimeout"`
 }
 
+type Parallelism struct {
+	// +kubebuilder:default:=1
+	// PP specifies the number of pods to create for the multi-node NIMService.
+	// +kubebuilder:validation:Minimum=1
+	PP int `json:"pp,omitempty"`
```
varunrsekar:
```suggestion
	Pipeline *uint32 `json:"pp,omitempty"`
```
can you call this `Pipeline`? PP doesnt sound good

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -158,6 +153,13 @@ type MultiNodeMPIConfig struct {
 	MPIStartTimeout int `json:"mpiStartTimeout"`
 }
 
+type Parallelism struct {
+	// +kubebuilder:default:=1
```
varunrsekar, line 157: we shouldnt have defaults here. This wouldnt work for multi-node anyway

internal/shared/nimservice.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		gpuQuantity, ok := nimService.Spec.Resources.Requests["nvidia.com/gpu"]
+		if !ok {
+			gpuQuantity, ok = nimService.Spec.Resources.Limits["nvidia.com/gpu"]
+			if !ok {
+				return 0, fmt.Errorf("GPU resource requests/limits not specified for NIMService %s in namespace %s", nimService.GetName(), nimService.GetNamespace())
+			}
+		}
+		return int(gpuQuantity.Value()), nil
+	} else {
```
varunrsekar, line 45: style nit: you dont need an else block here.

internal/shared/resourceclaims.go
```diff
@@ -235,3 +237,142 @@ func GetDRADeviceCELExpressions(device appsv1alpha1.DRADeviceSpec) ([]string, er
...
+// Constants for GPU device detection.
+const (
+	// NVIDIA identifiers used to detect GPU devices.
+	NVIDIAGPUComIdentifier = "gpu.nvidia.com"
+)
+
+// isNVIDIAGPU checks if a device class represents an NVIDIA GPU.
+func isNVIDIAGPU(ctx context.Context, client client.Client, deviceClassName string) (bool, error) {
+	if deviceClassName == "" {
+		return false, nil
```
varunrsekar, line 250: return error here. we shouldnt have an empty deviceClassName input

internal/shared/resourceclaims.go
```diff
@@ -235,3 +237,142 @@ func GetDRADeviceCELExpressions(device appsv1alpha1.DRADeviceSpec) ([]string, er
...
+	}
+
+	var dc resourcev1beta2.DeviceClass
+	if err := client.Get(ctx, types.NamespacedName{Name: deviceClassName}, &dc); err != nil {
+		return false, fmt.Errorf("failed to get device class %s: %w", deviceClassName, err)
+	}
+
+	hint := strings.ToLower(fmt.Sprint(dc.Spec))
+	return strings.Contains(hint, NVIDIAGPUComIdentifier), nil
```
varunrsekar, line 259: this is kinda flaky because it may just contain the term anywhere in the spec. Its better to be safe and iterating over all the selectors and looking only at the expressions.

As a follow-up we can also be careful in only checking if it contains `device.driver == "gpu.nvidia.com"` OR has `device.attributes["gpu.nvidia.com"]` OR has `device.capacity["gpu.nvidia.com"]`

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -385,14 +385,64 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
-func validateResourcesConfiguration(resources *corev1.ResourceRequirements, fldPath *field.Path) field.ErrorList {
+func validateResourcesConfiguration(spec *appsv1alpha1.NIMServiceSpec, fldPath *field.Path) field.ErrorList {
 	errList := field.ErrorList{}
-	if resources != nil {
-		if resources.Claims != nil || len(resources.Claims) != 0 {
-			errList = append(errList, field.Forbidden(fldPath.Child("claims"), "must be empty"))
+
+	// Validate that Claims must be empty
+	if spec.Resources != nil && len(spec.Resources.Claims) > 0 {
+		errList = append(errList, field.Forbidden(fldPath.Child("claims"), "must be empty"))
```
varunrsekar, line 393:
```suggestion
		errList = append(errList, field.Forbidden(fldPath.Child("resources").Child("claims"), "must be empty"))
```

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -385,14 +385,64 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
+	// Validate GPU requirements when MultiNode is enabled
+	if spec.MultiNode != nil {
+		errList = append(errList, validateGPURequirements(spec, fldPath)...)
+	}
+
+	return errList
+}
+
+// validateGPURequirements ensures that GPU resources are properly configured for MultiNode deployments.
+func validateGPURequirements(spec *appsv1alpha1.NIMServiceSpec, fldPath *field.Path) field.ErrorList {
```
varunrsekar, line 405: nit: call this `validateMultiNodeGPURequirements`?

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -385,14 +385,64 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
+		if !hasRequests && !hasLimits {
+			errList = append(errList, field.Required(fldPath.Child("resources"), "Either GPU requests or limits must be specified for MultiNode deployments"))
+			return errList
+		}
+
+		// Validate GPU requests if specified
+		if hasRequests {
+			gpuRequests := spec.Resources.Requests[gpuResourceName]
+			if gpuRequests.IsZero() || gpuRequests.Value() <= 0 {
+				errList = append(errList, field.Invalid(fldPath.Child("requests").Child("nvidia.com/gpu"), gpuRequests.String(), "must be greater than 0"))
```
varunrsekar, line 430:
```suggestion
				errList = append(errList, field.Invalid(fldPath.Child("requests").Key("nvidia.com/gpu"), gpuRequests.String(), "must be greater than 0"))
```

internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go
```diff
@@ -385,14 +385,64 @@ func validateScaleConfiguration(scale *appsv1alpha1.Autoscaling, fldPath *field.
...
+			}
 		}
+
+		// Validate GPU limits if specified
+		if hasLimits {
+			gpuLimits := spec.Resources.Limits[gpuResourceName]
+			if gpuLimits.IsZero() || gpuLimits.Value() <= 0 {
+				errList = append(errList, field.Invalid(fldPath.Child("limits").Child("nvidia.com/gpu"), gpuLimits.String(), "must be greater than 0"))
```
varunrsekar, line 438:
```suggestion
				errList = append(errList, field.Invalid(fldPath.Child("limits").Key("nvidia.com/gpu"), gpuLimits.String(), "must be greater than 0"))
```

visheshtanksale:
Closing this. Follow with webhook validations for next release

PR #642, opened 2025-08-29
Revert HTTPRoutes

shengnuo:


bundle/manifests/k8s-nim-operator.clusterserviceversion.yaml
```diff
@@ -1334,29 +1334,6 @@ spec:
...
-              - httproutes
-              verbs:
-              - create
-              - delete
-              - get
-              - list
-              - patch
-              - update
-              - watch
-            - apiGroups:
```
shivamerla, line 1349: DRA RBAC must be retained

internal/controller/nimservice_controller.go
```diff
@@ -294,60 +292,33 @@ func (r *NIMServiceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-
-	nimServiceBuilder, err = k8sutil.ControllerOwnsIfCRDExists(
-		r.discoveryClient,
-		nimServiceBuilder,
-		resourcev1beta2.SchemeGroupVersion.WithResource("resourceclaimtemplates"),
-		&resourcev1beta2.ResourceClaimTemplate{},
-	)
-	if err != nil {
-		return err
+	if resourceClaimCRDExists {
```
shivamerla, line 299: similarly we need to add for resourceclaimtemplates too, i prefer to keep the util method as-is and just revert the httproute checks.

PR #650, opened 2025-09-03
Updating parallelism fields

visheshtanksale:


varunrsekar:
> > Hey thanks for doing this, this is a lot simpler change.
> > Can you add validations?
> > 
> > * if spec.resources is provided, we'll need to validate that it matches spec.multiNode.parallelism.tensor
> > * spec.multiNode.parallelism.pipeline is required if multi-node
> > * spec.multiNode.parallelism.tensor is required if multi-node and spec.draResources is provided
> > 
> > Also, GPU count can still be inferred from spec.resources if spec.multiNode.parallelism.tensor is not provided. Would it make sense to implement that and just omit detection for DRA?
> 
> Decided to make only API changes as we are late into release cycle. Will add these in the next release

Okay sounds good

PR #657, opened 2025-09-04
Updating multi LLM samples without precaching

visheshtanksale:


config/samples/nim/serving/standalone/no-precaching/multi-llm-hf.yaml
```diff
@@ -12,10 +12,12 @@ spec:
     pullPolicy: IfNotPresent
     pullSecrets:
       - ngc-secret
-  authSecret: hf-secret # with HF_TOKEN set
+  authSecret: ngc-api-secret # with NGC_API_KEY set
   env:
     - name: NIM_MODEL_NAME
       value: hf://meta-llama/Llama-3.2-1B-Instruct
+    - name: HF_TOKEN
```
shivamerla, line 19: This will need to be part of secret as well.

visheshtanksale: Yes thats the fix. But this PR is just to fix the sample for v3.0.0 release.

PR #659, opened 2025-09-09
Using secret names and custom issuer with cert-manager

aryangorwade:
Implemented webhook signing configuration in a similar manner to DRA driver:

```
  admissionController:
    # Enable the admission controller. 
    # Note: cert-manager must be installed beforehand, as it is required to generate the TLS certificates.
    enabled: true    # TLS certificate configuration
    tls:
      # Certificate management mode: "cert-manager" or "secret"
      # - "cert-manager": Use cert-manager to automatically generate and manage certificates
      # - "secret": Use a user-provided secret containing tls.crt and tls.key
      mode: "secret"
      certManager:
        # Issuer type: "selfsigned", "clusterissuer", or "issuer"
        issuerType: "selfsigned"
        # Issuer name (required when issuerType is "clusterissuer" or "issuer")
        issuerName: ""
        # Additional DNS names for the certificate
        dnsNames: []
      secret:
        # Name of the secret containing tls.crt and tls.key
        name: ""
        # Base64-encoded CA certificate bundle for validating the webhook's TLS certificate (base64 encoded)
        # Required when using secret mode.
        # Note: Only include intermediate CA certificates, not root CA certificates
        caBundle:
```

This change introduces a new Helm chart value to supply a CA bundle (e.g., from ca.crt) directly into the ValidatingWebhookConfiguration when cert-manager is not used. Secret must be created with tls.key and tls.crt.
This allows clusters without automated CA injection to configure the webhook manually and ensure proper TLS verification.


deployments/helm/k8s-nim-operator/values.yaml
```diff
@@ -56,8 +56,26 @@ operator:
...
     # Enable the admission controller. 
     # Note: cert-manager must be installed beforehand, as it is required to generate the TLS certificates.
-    enabled: false
-
+    enabled: true    # TLS certificate configuration
+    tls:
+      # Certificate management mode: "cert-manager" or "secret"
+      # - "cert-manager": Use cert-manager to automatically generate and manage certificates
+      # - "secret": Use a user-provided secret containing tls.crt and tls.key
+      mode: "secret"
```
shivamerla, line 64: default can be `cert-manager` to avoid users having to generate the key-pair and create a secret.

shivamerla:
Thanks @aryangorwade this looks good to me.

deployments/helm/k8s-nim-operator/values.yaml
aryangorwade: Updated.

PR #668, opened 2025-09-11
Integrate GRPCRoute into NIMService

shengnuo:
This MR adds integrates NIMServices with GRPCRoute.

- GRPCRoutes can be enabled via `.spec.router.gateway.grpcRoutesEnabled`
```yaml
  router:
    gateway:
      namespace: nemo
      name: istio-gateway
      grpcRoutesEnabled: true
    hostDomainName: demo.nvidia.example.com
```

- GRPC Port is specified under `.spec.expose.service.grpcPort`. If it's not specified, the default `50051` port is used.
- If HTTPRoute is enabled along with GRPCRoute, then HTTPRoute and GRPCRoute objects will be created with the same hostname specified under `.spec.router.hostname`

**Sample Istio Gateway (HTTP + gRPC on same hostname)**
```
apiVersion: gateway.networking.k8s.io/v1
kind: Gateway
metadata:
  name: istio-gateway
  namespace: nemo
spec:
  gatewayClassName: istio
  listeners:
  - allowedRoutes:
      namespaces:
        from: All
      kinds:
      - kind: HTTPRoute
    hostname: riva-tts.nemo.demo.nvidia.example.com
    name: http
    port: 80
    protocol: HTTP
  - allowedRoutes:
      namespaces:
        from: All
      kinds:
      - kind: GRPCRoute
    hostname: riva-tts.nemo.demo.nvidia.example.com
    name: grpc
    port: 50051
    protocol: HTTP
```

### DIFF
--- a/api/apps/v1alpha1/nimservice_types.go
+++ b/api/apps/v1alpha1/nimservice_types.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"strings"
 
 	kserveconstants "github.com/kserve/kserve/pkg/constants"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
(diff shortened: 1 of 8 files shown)


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -1350,6 +1366,16 @@ func (n *NIMService) GetHTTPRouteParams() *rendertypes.HTTPRouteParams {
 	return params
 }
 
+func (n *NIMService) GetGRPCRouteParams() *rendertypes.GRPCRouteParams {
```
shivamerla, line 1369: we seem to have missed setting annotations here

shivamerla (dismissed):

shivamerla (approved):

PR #673, opened 2025-09-24
Fixing e2e

visheshtanksale:
- Kept Holodeck version consistent with GPU Operator
- Added Calico installation
- Removed installation on Admission webhook on Helm Chart
- Updated NIMs version that is being installed
- Use bigger AWS instance

To Do
- Update Holodeck
- Add more coverage
- Run Gather All script on failure


shivamerla (approved):
LGTM

test/e2e/infra/aws.yml
```diff
@@ -29,4 +29,6 @@ spec:
   kubernetes:
     install: true
     installer: kubeadm
-    version: v1.32.1
+    version: v1.32.3
```
shivamerla, line 32: If we can bump this to 1.34, then we can test with DRA as well.

PR #675, opened 2025-09-29
Add support for building nim-operator on nvidia hosted runners and using custom GOPROXY

shivamerla:


.github/workflows/golang.yml
```diff
@@ -48,13 +45,41 @@ jobs:
         version: latest
         args: -v --timeout 5m
         skip-cache: true
+    - name: Setup JFrog CLI
+      id: jfrog
+      uses: jfrog/setup-jfrog-cli@v4
+      env:
+        JF_URL: https://${{ secrets.OIDC_ARTIFACTORY_ENDPOINT }}/
+        OIDC_USER: ${{ steps.jfrog.outputs.oidc-user }}
```
visheshtanksale, line 53: Remove this `steps.jfrog.outputs.oidc-user` will only be available after jfrog runs

.github/workflows/golang.yml
```diff
@@ -67,10 +92,32 @@ jobs:
       uses: actions/setup-go@v5
       with:
         go-version: ${{ env.GOLANG_VERSION }}
```
visheshtanksale, line 94: There is `cache: true` that might be able to save some time on the actions run below this

PR #678, opened 2025-09-30
Updating container images for bitnami charts to bitnamilegacy repo

visheshtanksale:


shivamerla (approved):
thanks for taking care of this.

PR #681, opened 2025-10-07
While performing update retain current replicas of deployment set by the HPA

shivamerla:
This avoids the operator from resetting the replicas back to 1 (default)  on updates and let HPA manage them.


internal/controller/nemo_datastore_controller.go
```diff
@@ -493,20 +493,9 @@ func (r *NemoDatastoreReconciler) reconcilePVC(ctx context.Context, nemoDatastor
...
-	if err != nil && !errors.IsNotFound(err) {
-		logger.Error(err, fmt.Sprintf("Error is not NotFound for %s: %v", obj.GetObjectKind(), err))
-		return err
-	}
-	// Don't do anything if CR is unchanged.
-	if err == nil && !utils.IsParentSpecChanged(obj, utils.DeepHashObject(nemoDatastore.Spec)) {
-		return nil
-	}
-
 	resource, err := renderFunc()
```
varunrsekar, line 496: I remember we needed this check before the `renderFunc` so that any failures in `renderFunc` don't simply mark the CR as failed on nim-operator restarts.

internal/controller/nemo_datastore_controller.go
```diff
@@ -531,6 +520,31 @@ func (r *NemoDatastoreReconciler) renderAndSyncResource(ctx context.Context, nem
...
+
+	// If we found the object and autoscaling is enabled on the Datastore,
+	// copy the current replicas from the existing object into the desired (resource),
+	// so we don't fight the HPA (or external scaler) on each reconcile.
+	if err == nil && nemoDatastore.IsAutoScalingEnabled() {
+		switch curr := obj.(type) {
+		case *appsv1.Deployment:
+			if desired, ok := resource.(*appsv1.Deployment); ok && curr.Spec.Replicas != nil {
+				replicas := *curr.Spec.Replicas
+				desired.Spec.Replicas = &replicas
```
varunrsekar, line 538: I'm curious why we need to preserve the autoscaler replicas. If a deployment is rolled out because of a spec change, wouldnt all the metrics reset and its better we start back from `1`?

We also have a lack of understanding of who changed the replicas:
1. the user manually updated it (even w/ HPA).
2. the user didnt touch it and only the HPA updated it.

Do we care about preserving it in both cases even if the NIMService itself has only 1 replica in its spec.?

PR #698, opened 2025-10-31
Adding HF token env var when using multi llm NIM without caching

visheshtanksale:


internal/controller/platform/standalone/nimservice.go
```diff
@@ -462,6 +462,26 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
 				Value: utils.DefaultModelStorePath,
 			}}, deploymentParams.Env)
 		}
+		// If NIMCache is not provided, and the model is from Hugging Face, add the HF_TOKEN to the environment variables
```
shivamerla: Should we just include all env vars from the secret by default?
Example [here](https://github.com/NVIDIA/k8s-nim-operator/blob/main/api/apps/v1alpha1/nimcache_types.go#L221)
.
Maybe we can add a helper that returns all env vars from a secret as a string.

visheshtanksale: The current implementation of setting the env by explicitly  selecting  `NGC_API_KEY` from the secrets ensures that the pod is not scheduled if the secret does not have `NGC_API_KEY` key. I feel that is a better user experience than the pod getting scheduled and then going through the logs to figure out that the secret was not created as expected.
Also, setting `NGC_API_KEY` is the majority of use cases of the NIM Operator.

visheshtanksale: Ok addressing it.

PR #711, opened 2025-11-05
Improving the condition for adding HF_TOKEN on NIM Service

visheshtanksale:


internal/controller/platform/standalone/nimservice.go
```diff
@@ -464,28 +464,25 @@ func (r *NIMServiceReconciler) reconcileNIMService(ctx context.Context, nimServi
...
-								SecretKeyRef: &corev1.SecretKeySelector{
-									LocalObjectReference: corev1.LocalObjectReference{
-										Name: nimService.Spec.AuthSecret,
-									},
-									Key: "HF_TOKEN",
-								},
+
+		// If NIMCache or NIMService is a Hugging Face Multi-LLM NIM, add the HF_TOKEN to the environment variables
+		if nimCache.IsHFMultiLLM() || nimService.IsHFMultiLLM() {
+			deploymentParams.Env = utils.RemoveEnvVar(deploymentParams.Env, "NGC_API_KEY")
```
varunrsekar: Can you make all these env vars a const? We seem to be referencing them in multiple places

visheshtanksale: Fixed

api/apps/v1alpha1/nimcache_types.go
```diff
@@ -331,6 +331,11 @@ func (n *NIMCache) IsUniversalNIM() bool {
 	return false
 }
 
+// IsHFMultiLLM returns true if the NIMCache is for a Hugging Face Multi-LLM NIM.
+func (n *NIMCache) IsHFModel() bool {
+	return n.Spec.Source.HF != nil
```
varunrsekar: Do we need to worry about nemo-datastore here? `n.Spec.Source.Datastore.AuthSecret` also requires a HF_TOKEN secret

shivamerla (dismissed):

PR #712, opened 2025-11-06
Support computedomain for multi-node NIMServices

varunrsekar:
* TODO: Resolve NVIDIA/k8s-dra-driver-gpu dependency
* TODO: Resolve kserve not supporting k8s.io/* dependency version > v1.33


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -167,6 +172,8 @@ type MultiNodeMPIConfig struct {
 	MPIStartTimeout int `json:"mpiStartTimeout"`
 }
 
+type NIMServiceComputeDomainSpec struct{}
```
visheshtanksale: Should be called `ComputeDomainSpec`

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -1817,6 +1834,31 @@ func (n *NIMService) GetMultiNodePipelineParallelism() uint32 {
 	return 0
```
visheshtanksale: No part of your change, but can you please update it?
`GetMultiNodePipelineParallelism()` is looking for `n.Spec.MultiNode.Parallelism.Tensor`nil check instead of `n.Spec.MultiNode.Parallelism.Pipeline`

manifests/computedomain.yaml
```diff
@@ -0,0 +1,17 @@
...
+  namespace: {{ .Namespace }}  
+  {{- if .Labels }}
+  labels:
+    {{- .Labels | yaml | nindent 4 }}
+  {{- end }}
+  {{- if .Annotations }}
+  annotations:
+    {{- .Annotations | yaml | nindent 4 }}
+  {{- end }}
+  spec:
```
shivamerla: extra indentation here for the spec field.

PR #732, opened 2025-12-10
Use v1 DRA APIs

varunrsekar:
DRA is GA in k8s v1.34! This change bumps up the consumed DRA APIs from v1beta2 -> v1 api version.

Changes:
1. Bump go mod dependencies for k8s libs to v0.34
2. Pin kserve libs to a commit in the master branch that supports k8s v0.34 libs (https://github.com/kserve/kserve/tree/1cb39eba0f1a/) - this is a temporary change until kserve v0.17.0 is released.
3. Tweaks in all kserve-related code to support the above version

Note:
1. With this change, NIM Operator will no longer support k8s versions below 1.34 for consuming DRA.
2. NIM Operator will no longer support DRA resourceclaims/resourceclaimtemplates of api version v1beta2

Testing:
1. Verified w/ DRA and w/o DRA with k8s cluster version v1.34
3. Verified w/o DRA with k8s cluster version v1.33

### DIFF
--- a/internal/controller/platform/kserve/nimservice.go
+++ b/internal/controller/platform/kserve/nimservice.go
@@ -463,6 +463,11 @@ func (r *NIMServiceReconciler) renderAndSyncInferenceService(ctx context.Context
 	isvcParams.Annotations[kserveconstants.EnableMetricAggregation] = "true"
 	isvcParams.Annotations[kserveconstants.SetPrometheusAnnotation] = "true"
 
+	// Ensure deployment mode annotation is always set
+	if _, ok := isvcParams.Annotations[utils.KServeDeploymentModeAnnotationKey]; !ok {
+		isvcParams.Annotations[kserveconstants.DeploymentMode] = string(deploymentMode)
+	}
+
 	// Sync ingress
 	if !nimService.IsIngressEnabled() {
 		isvcParams.Labels[kserveconstants.NetworkVisibility] = kserveconstants.ClusterLocalVisibility


xieshenzh:
These test cases should be updated, since `Standard` is now the default mode when the annotation is absent:
https://github.com/NVIDIA/k8s-nim-operator/blob/f628f6cf905082e0ce6a4db47586a1731c3abd60/internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper_test.go#L1002-L1080

xieshenzh:
These examples should be updated as well, since `Serverless` is not supposed to be the default mode when the annotation is [absent](https://github.com/NVIDIA/k8s-nim-operator/blob/f628f6cf905082e0ce6a4db47586a1731c3abd60/vendor/github.com/kserve/kserve/pkg/apis/serving/v1beta1/configmap.go#L349):
https://github.com/NVIDIA/k8s-nim-operator/tree/f628f6cf905082e0ce6a4db47586a1731c3abd60/config/samples/nim/serving/kserve

varunrsekar:
> These test cases should be updated, since the default mode is supposed to be read from a [configmap](https://github.com/NVIDIA/k8s-nim-operator/blob/f628f6cf905082e0ce6a4db47586a1731c3abd60/vendor/github.com/kserve/kserve/pkg/apis/serving/v1beta1/configmap.go#L349) when the annotation is absent:

@xieshenzh Do you mean that we need to setup a configmap with default deployment mode as KNative in these tests?

What's the expected behavior if the configmap is empty and we don't set any annotation in the ISvc? this [testcase](https://github.com/kserve/kserve/blob/e400006952af8913ca5ee7cc9aa771288948b63a/pkg/apis/serving/v1beta1/configmap_test.go#L467) indicates the configmap being empty is valid.

varunrsekar:
> I think the expected behavior if the configmap is empty is to use the standard mode.
If to use standard mode, the reconciler of NIMService should set the annotation explicitly when creating the inferenceservice, in order to avoid any potential issues from the kserve controller.

I see [here](https://github.com/kserve/kserve/blob/e400006952af8913ca5ee7cc9aa771288948b63a/pkg/controller/v1beta1/inferenceservice/components/predictor.go#L189) that the predictor reconciler defaults to knative even though we're expected to [exit early](https://github.com/kserve/kserve/blob/e400006952af8913ca5ee7cc9aa771288948b63a/pkg/controller/v1beta1/inferenceservice/controller.go#L214) in case of missing knative CRDs

> There is a constant for the default deployment mode, which is [standard](https://github.com/kserve/kserve/blob/e400006952af8913ca5ee7cc9aa771288948b63a/pkg/constants/constants.go#L468), but unfortunately it is not used anywhere in the code.

[isvcutils.GetDeploymentMode](https://github.com/kserve/kserve/blob/master/pkg/controller/v1beta1/inferenceservice/utils/utils.go#L244) seems to provide the default deployment. Isn't this behavior sufficient? Or do you think we should set this as an annotation explicitly from the NIMService?

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -1638,7 +1638,7 @@ func (n *NIMService) GetInferenceServiceParams(
 	delete(params.PodAnnotations, utils.NvidiaAnnotationParentSpecHashKey)
 
 	// Set template spec
-	if !n.IsAutoScalingEnabled() || deploymentMode != kserveconstants.RawDeployment {
+	if !n.IsAutoScalingEnabled() || (deploymentMode != kserveconstants.LegacyRawDeployment && deploymentMode != kserveconstants.Standard) {
```
shivamerla: nit: can we introduce a helper method for checking these two modes?

shivamerla (dismissed):
LGTM

api/apps/v1alpha1/nimservice_types.go
varunrsekar: I've cleaned up the commits and added an util `IsKServeStandardDeploymentMode`

PR #734, opened 2025-12-17
Fixing e2e test

visheshtanksale:
- NGC API secret had expired. Updated it on GH
- Update NIM CR clean up logic to be only executed when test is successful. This allows the corresponds pod to be part of test logs archive
- Update HF token

### DIFF
--- a/test/e2e/nim-operator_test.go
+++ b/test/e2e/nim-operator_test.go
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
-	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
 
 	helm "github.com/mittwald/go-helm-client"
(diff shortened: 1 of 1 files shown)


test/e2e/nim-operator_test.go
```diff
@@ -159,7 +162,7 @@ var _ = Describe("NIM Operator", Ordered, func() {
 
 		AfterEach(func() {
 			// Clean up
-			if !CurrentSpecReport().Failed() {
+			if CurrentSpecReport().State.Is(types.SpecStatePassed) {
```
varunrsekar: this change isnt required right? `Failed()` covers the following states:
* failed
* aborted
* panicked
* interrupted
* timedout

the remaining ones are:
* invalid
* pending
* skipped
* passed

We should just use `!CurrentSpecReport().Failed()` everywhere.

varunrsekar (commented):
LGTM. Just 1 comment...

test/e2e/nim-operator_test.go
visheshtanksale: Fixed it!!

Issue #710, opened 2025-11-05
Add support for scaling down NIMServices to 0 replicas


tgdfool2:
In our infrastructure, we are testing different models and we often need to scale some models down to 0 replicas to free up capacity for new ones, while wanting to quickly be able to restart some replicas if needed.

At the moment, the minimum amount of replicas allowed for a NIMService is 1; it would be nice to be able to scale it down to 0 as well.

With older NIM Operator versions, we could work around this issue by setting the number of replicas to 0 at the Deployment level (`k scale deployment mistral-small-3-2-24b-instruct-2506 --replicas 0`) but, with newer NIM Operator versions, this produces quite a lot of noise in the logs of the NIM Operator:
_(every 5 seconds)_
```
nim-operator-k8s-nim-operator-5fcd5f57b-jvf4k manager {"level":"info","ts":"2025-11-04T07:48:46Z","msg":"Reconciling","controller":"nimservice","controllerGroup":"apps.nvidia.com","controllerKind":"NIMService","NIMService":{"name":"mistral-small-3-2-24b-instruct-2506","namespace":"test-ns"},"namespace":"test-ns","name":"mistral-small-3-2-24b-instruct-2506","reconcileID":"5cc948e0-5599-4696-bfd8-1ce32cf6dc1e","NIMService":"mistral-small-3-2-24b-instruct-2506"}
nim-operator-k8s-nim-operator-5fcd5f57b-jvf4k manager {"level":"info","ts":"2025-11-04T07:48:46Z","logger":"controllers.NIMService","msg":"Reconciling NIMService instance","nimservice":"mistral-small-3-2-24b-instruct-2506"}
[...]

PR #738, opened 2026-01-22
Updating the replicas for kserve

visheshtanksale:
- Fixing https://github.com/NVIDIA/k8s-nim-operator/issues/710
- Not adding replicas on InferenceService for Knative/serverless deployment mode

### DIFF
--- a/internal/controller/platform/standalone/nimservice.go
+++ b/internal/controller/platform/standalone/nimservice.go
@@ -909,6 +909,9 @@ func (r *NIMServiceReconciler) isLeaderWorkerSetReady(ctx context.Context, nimSe
 		}
 		return "", false, err
 	}
+	if leaderWorkerSet.Spec.Replicas == nil || *leaderWorkerSet.Spec.Replicas == 0 {
+		return fmt.Sprintf("leaderworkerset %q is scaled down", leaderWorkerSet.Name), false, nil
+	}
 
 	for _, cond := range leaderWorkerSet.Status.Conditions {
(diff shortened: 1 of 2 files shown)


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -123,7 +123,7 @@ type NIMServiceSpec struct {
 	Scale          Autoscaling   `json:"scale,omitempty"`
 	SchedulerName  string        `json:"schedulerName,omitempty"`
 	Metrics        Metrics       `json:"metrics,omitempty"`
-	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Minimum=0
```
varunrsekar: For non-kserve NIMs, is it expected to be able to deploy it with 0 replicas?

shivamerla: Based on the issue linked looks like the user want to spin it down to `0` to optimize GPU usage as we don't have scale-to-zero support yet. @visheshtanksale if we want to support this then we need to avoid trying to set model status as well and set the status as `NotReady` or introduce a new state.

shivamerla: @visheshtanksale can you double check this for LWS and InferenceService as well. Does `isLeaderWorkerSetReady` and `isInferenceServiceReady` return the right state if replicas set to `0`, if they return ready by any chance then we end up attempting to set the model status

shivamerla (commented):
LGTM

shivamerla (approved):

PR #752, opened 2026-02-04
Expose Sidecar and Init Container functionality

soluwalana:
This MR exposes Sidecar and Init Containers for the NIMService and NIMCache CRDS.

This allows end users to enable the following types of functionality:

Secret Injection in an Init Container
Proxy Container for other services in the Kubernetes
Proxy Container for Additional flexibility with Agents
Sidecar Container for Loading LoRA adapters from different locations (HF, NGC, NMP Files Service)

### DIFF
--- a/bundle/manifests/apps.nvidia.com_nimcaches.yaml
+++ b/bundle/manifests/apps.nvidia.com_nimcaches.yaml
@@ -233,16 +233,24 @@ spec:
                 type: integer
               initContainers:
                 items:
+                  description: NIMContainerSpec defines the specification for a container
+                    within a NIM workload.
                   properties:
                     args:
+                      description: Args are the arguments passed to the container
+                        command at runtime.
                       items:
                         type: string
                       type: array
                     command:
+                      description: Command is the entrypoint array. If not specified,
+                        the image's default entrypoint is used.
                       items:
                         type: string
                       type: array
                     env:
+                      description: Env is the list of environment variables to set
+                        in the container.
                       items:
                         description: EnvVar represents an environment variable present
                           in a Container.
@@ -400,7 +408,7 @@ spec:
                         type: object
                       type: array
                     image:
-                      description: Image defines image attributes.
+                      description: Image specifies the container image to run.
                       properties:
                         pullPolicy:
                           type: string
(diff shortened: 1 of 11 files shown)


api/apps/v1alpha1/nimcache_types.go
```diff
@@ -401,25 +403,45 @@ func (n *NIMCache) GetEnvWithProxy() []corev1.EnvVar {
...
+			image = n.Spec.Source.HF.ModelPuller
 		}
-		return initContainerList
+		initContainers = append(initContainers, corev1.Container{
+			Name:            "update-ca-certificates",
+			Image:           image,
+			Command:         k8sutil.GetUpdateCaCertInitContainerCommand(),
+			SecurityContext: k8sutil.GetUpdateCaCertInitContainerSecurityContext(),
+			VolumeMounts:    k8sutil.GetUpdateCaCertInitContainerVolumeMounts(),
+		})
```
shivamerla: Better to prepend this container before others, so that certs are updated first.

soluwalana: For sure, will update.

soluwalana:
Yeah the complexity and functionality constraints of adding it to a ConfigMap is pretty high: 
* An extra step required post installation for config map creation either at the operator or workload namespace level.
* Whenever the config map changes (version bumps etc), we have to rollout those changes. A config-hash annotation can be used for this.
* Debugging issues will be hard.
* Difficulty of type checking the ConfigMap object. 
* No current ConfigMap loading anywhere in the code base
* Inability to change Env Variables (required for sidecar proxies) via the config map


@varunrsekar This PR is the simplest change set necessary to accomplish the proxy and initialization requirements we have. 

The Env Vars being different between each NimService's sidecar and init container (for injecting principal information for example) precludes the ability to use a config map as the source for the definition.

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -89,6 +89,15 @@ const (
 	PlatformTypeKServe PlatformType = "kserve"
 )
 
+type NIMContainerSpec struct {
```
visheshtanksale: Please move this to `common_types.go`

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -853,8 +873,8 @@ func (n *NIMService) GetVolumeMounts(modelPVC *PersistentVolumeClaim) []corev1.V
 			SubPath:   subPath,
 		},
 		{
-			Name:      "dshm",
-			MountPath: "/dev/shm",
+			Name:      "scratch",
```
visheshtanksale: Why is the dshm VolumeMount being removed?

PR #755, opened 2026-02-11
fix: use spec.replicas for minReplicas when autoscaling is disabled

TomerFi:
## Problem

When autoscaling is disabled under Standard/RawDeployment mode, `GetInferenceServiceParams()` hardcoded `minReplicas` and `maxReplicas` to 0. This caused the Dashboard to display "0 replicas" even though the deployment was running correctly.

## Fix

Use `GetReplicas()` to read `spec.replicas` from the `NIMService` instead, so the `InferenceService` correctly reflects the configured replica count.

## Testing

- Unit tests added for the updated `GetInferenceServiceParams()` behavior


TomerFi:
@mtalvi Are we adding `security.opendatahub.io/enable-auth` to the NIMService's `metadata.annotations`? If so, can we move it to `spec.annotations`? Thanks.

PR #761, opened 2026-02-16
Add sidecar containers to NeMo workloads

### LINKED ISSUE #760: Support IAM authentication to PostgreSQL databases
Currently Nvidia Kubernetes NIM Operator doesn't support Google Cloud IAM authentication to Google Cloud SQL (for example), and since you cannot run a sidecar container, you also cannot run the [Google Cloud SQL Proxy](https://github.com/GoogleCloudPlatform/cloud-sql-proxy/tree/main) sidecar container to use IAM authentication.

**Describe the solution you'd like**
I would like the ability to create a sidecar container for all of the custom resources which use PostgreSQL to run the Google Cloud SQL Proxy as a sidecar so that IAM authentication to Google Cloud SQL can be used.

**Describe alternatives you've considered**
Plain authentication instead IAM authentication, but this is less secure and has the overhead of rotating passwords frequently being necessary.

benglewis:
# Changes

  - Added a spec.sidecarContainers slice to each NeMo CRD so operators can declaratively inject
    additional containers (e.g., IAM proxy) and plumbed that field through the deployment/
    statefulset params in api/apps/v1alpha1/nemo_*_types.go (:61–:77, :498, :684, :733, :701) plus
    the autogenerated deepcopy logic at api/apps/v1alpha1/
    zz_generated.deepcopy.go:1851, :2076, :2285, :2473, :2671.
  - Taught the shared render layer to carry sidecars (internal/render/types/types.go:86, :160) and
    appended them to the pod templates used by both Deployment and StatefulSet manifests
    (manifests/deployment.yaml:118, manifests/statefulset.yaml:73).
  - Regenerated all CRD outputs (config, Helm, bundle) so the schema knows about the new field
    (config/crd/bases/apps.nvidia.com_nemo*.yaml:2476/2567/2612/2878/2600 and matching Helm/bundle
    manifests).

# Testing
  - go test ./api/apps/v1alpha1 ./internal/render — fails during setup because vendor/k8s.io/api/
    extensions/v1beta1/zz_generated.validations.go:29-30 imports k8s.io/apimachinery/pkg/api/
    {safe,validate}, which can’t be resolved with -mod=vendor.
  - go test ./internal/controller -run 'Nemo(Datastore|Entitystore|Evaluator|Customizer|Guardrail)'
    — same vendor import errors as above.

Closes #760 .

Disclaimer: All changes were written by Codex

### DIFF
--- a/api/apps/v1alpha1/nemo_customizer_types.go
+++ b/api/apps/v1alpha1/nemo_customizer_types.go
@@ -70,15 +70,16 @@ const (
 // +kubebuilder:validation:XValidation:rule="!(has(self.expose.ingress) && has(self.expose.ingress.enabled) && self.expose.ingress.enabled && has(self.expose.router) && has(self.expose.router.ingress))", message=".spec.expose.ingress is deprecated, and will be removed in a future release. If .spec.expose.ingress is set, please do not set .spec.expose.router.ingress."
 // +kubebuilder:validation:XValidation:rule="!(has(self.scale) && has(self.scale.enabled) && self.scale.enabled && has(self.replicas))",message="spec.replicas cannot be set when spec.scale.enabled is true"
 type NemoCustomizerSpec struct {
-	Image        Image               `json:"image"`
-	Command      []string            `json:"command,omitempty"`
-	Args         []string            `json:"args,omitempty"`
-	Env          []corev1.EnvVar     `json:"env,omitempty"`
-	Labels       map[string]string   `json:"labels,omitempty"`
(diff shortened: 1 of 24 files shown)


rajatchopra:
Closing this PR.

Reasoning: A native implementation is already done here (https://github.com/NVIDIA/k8s-nim-operator/commit/1ee52054f46569793f2d5dc30b2bb5f0b4295d8e), which does not use corev1.ContainerSpec directly.

@benglewis If you think this PR is still needed and that it implements the sidecar feature in a better way, please feel free to re-open.

PR #764, opened 2026-02-18
fix: configure TLS and bearer auth for HTTPS model status checks

TomerFi:
## Problem

When token auth is enabled on a NIMService (via the `security.opendatahub.io/enable-auth: "true"` annotation, propagated from `spec.annotations` to the InferenceService), KServe injects a `kube-rbac-proxy` sidecar that:
1. Serves HTTPS on port 8443 using OpenShift service-serving certificates
2. Requires bearer token authentication (validated via SubjectAccessReview)
3. Proxies authenticated requests to the NIM container on port 8000

This changes the InferenceService's `Status.Address.URL` from `http://...:8000` to `https://...:8443`. The operator's status reconciler calls this URL to query `/v1/models`, but `doGetRequest()` used a default `http.Client` with no CA configuration and no auth headers. This caused two failures:

1. **TLS verification failed** — `x509: certificate signed by unknown authority` (the service-serving CA is not in the system trust store)
2. **Authentication failed** — `401 Unauthorized` (no bearer token provided for `kube-rbac-proxy`)

As a result, NIMService CRs never transition to `Ready`, even though the underlying deployment is fully functional.

Ref: https://issues.redhat.com/browse/NVPE-417

## Fix

For HTTPS URLs, `doGetRequest()` now:
- Loads the OpenShift service-serving CA bundle from a mounted ConfigMap (annotated with `service.beta.openshift.io/inject-cabundle: "true"`) to establish TLS trust
- Reads the operator's ServiceAccount token and includes it as a `Bearer` authorization header for `kube-rbac-proxy` authentication

For HTTP URLs (standalone mode without auth proxy), the default client is used unchanged.

## No impact on non-OpenShift / non-KServe deployments

Every new behavior is guarded by graceful fallbacks:
- The ConfigMap volume is `optional: true` — the pod starts normally if the ConfigMap doesn't exist
- If the CA bundle file is missing, a debug-level log is emitted and the default TLS config is used
- If the SA token file is missing, auth is skipped
- All new code paths are gated on `parsedURL.Scheme == "https"` — HTTP requests are completely unaffected

## Testing

- **Unit tests** added for `newTLSTransport()` (valid cert, missing file, invalid PEM, empty file) and `readBearerToken()` (valid token, whitespace trimming, missing file, empty file, whitespace-only file)
- `make lint` — 0 issues
- `make test` — all packages pass
- **Cluster validation** on OpenShift with RHOAI:
  - After deploying the TLS fix, the `x509: certificate signed by unknown authority` errors disappeared from the operator logs, replaced by `401 Unauthorized` — confirming TLS verification was resolved
  - After adding the bearer token fix, an existing NIMService that was stuck in a `401 Unauthorized` retry loop transitioned to `Ready: True` on its own through the operator's reconcile loop, with no manual restart needed

### DIFF
--- a/internal/nimmodels/models.go
+++ b/internal/nimmodels/models.go
@@ -111,18 +111,17 @@ func newTLSTransport(caBundlePath string) (*http.Transport, error) {
 }
 
 // readBearerToken reads a bearer token from the given file path.
-// Returns the token string and true if successful, or empty string and false
-// if the file doesn't exist or can't be read.
-func readBearerToken(tokenPath string) (string, bool) {
+// Returns the token string or an error if the file cannot be read or is empty.
+func readBearerToken(tokenPath string) (string, error) {
 	token, err := os.ReadFile(tokenPath)
 	if err != nil {
-		return "", false
+		return "", fmt.Errorf("reading token from %s: %w", tokenPath, err)
 	}
 	t := strings.TrimSpace(string(token))
 	if t == "" {
-		return "", false
+		return "", fmt.Errorf("token file %s is empty", tokenPath)
 	}
-	return t, true
+	return t, nil
 }
 
 func doGetRequest(ctx context.Context, requestURL string) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


visheshtanksale (dismissed):
Overall LGTM!. Minor comment

visheshtanksale (approved):
LGTM

PR #819, opened 2026-04-02
feat: add priorityclass to nimserivces

### LINKED ISSUE #731: Add support for priorityclasses
Hi,

we want to ensure that pods resulting from `NimServices` in certain namespaces always get scheduled, even if all GPUs are being used as a new `NimService` is added. We would like to use [PriorityClasses](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass) for that. 

As of now, I don't see a way of adding the required parameters to a `NimService`.

Thanks in advance!

crust3780:
fixes #731 

unfortunately i was not able to regenerate the bundled crds because i had issues with the operator sdk.


visheshtanksale:
Closing this. Being fixed here https://github.com/NVIDIA/k8s-nim-operator/pull/866

PR #833, opened 2026-04-28
Fix KServe namespace discovery 

xieshenzh:
Fixes an issue where creating a NIMService fails with:                                                                                                                                                   
 ```
admission webhook "vnimservice-v1alpha1.kb.io" denied the request:
  nimservice.spec: Internal error: failed to determine KServe deployment mode:                                                                                                                             
  failed to find the namespace of KServe Deployment  
```
The operator was looking for KServe's controller Deployment using label `app.kubernetes.io/name=kserve-controller-manager`, then using the namespace of the Deployment to retrieve the ConfigMap `inferenceservice-config`. But, the label was recently modified to `app.kubernetes.io/name=kserve` instead, causing the lookup to fail.

This PR:
  - Switch to the `control-plane=kserve-controller-manager` label, which is part of KServe Controller Deployment's `spec.selector.matchLabels` and is not supposed to be changed or overridden by kustomize.                           
  - Implement a two-tier strategy for locating the `inferenceservice-config` ConfigMap:
    - Primary: if the `KSERVE_NAMESPACE` env var is set, fetch the ConfigMap from that namespace directly. Return nil gracefully if not found.                                                               
    - Fallback: if `KSERVE_NAMESPACE` is unset, discover the namespace dynamically by listing Deployments with label `control-plane=kserve-controller-manager`. Expects exactly one match; errors with guidance to set `KSERVE_NAMESPACE` if multiple are found.                                                                                                                                                          

 Tests:       
  - Verify ConfigMap is fetched from `KSERVE_NAMESPACE` when set and ConfigMap exists                                                                                                                        
  - Verify nil is returned when `KSERVE_NAMESPACE` is set but ConfigMap does not exist                                                                                                                       
  - Verify dynamic discovery via `control-plane=kserve-controller-manager` label works when `KSERVE_NAMESPACE` is unset                                                                                        
  - Verify ConfigMap is fetched from the discovered namespace                                                                                                                                              
  - Verify nil is returned when ConfigMap does not exist in the discovered namespace                                                                                                                       
  - Verify error with `KSERVE_NAMESPACE` hint when no KServe controller deployment is found                                                                                                                  
  - Verify error with `KSERVE_NAMESPACE` hint when multiple KServe controller deployments are found                                                                                                          
  - Verify existing deployment mode detection tests pass with updated label

### DIFF
--- a/internal/controller/platform/kserve/nimservice_test.go
+++ b/internal/controller/platform/kserve/nimservice_test.go
@@ -410,10 +410,10 @@ var _ = Describe("NIMServiceReconciler for a KServe platform", func() {
 
 		kserveDeployment := &appsv1.Deployment{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:      utils.KServeControllerName,
+				Name:      "kserve-controller-manager",
 				Namespace: "default",
 				Labels: map[string]string{
-					"app.kubernetes.io/name": utils.KServeControllerName,
+					"control-plane": "kserve-controller-manager",
 				},
 			},
 			Spec: appsv1.DeploymentSpec{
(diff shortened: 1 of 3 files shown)


visheshtanksale (approved):
LGTM

PR #864, opened 2026-06-16
Add Labels field to Service struct

JunAr7112:
This MR is in response to this [issue](https://github.com/NVIDIA/k8s-nim-operator/issues/806). Essentially, the goal is to add an additional expose.service.labels field that would allow the user to select NIM generated services by their label.

SOLUTION:
In _type Service struct_, a labels field was added:
`Labels      map[string]string `json:"labels,omitempty"``

A helper function GetExposeServiceLabels() was defined. It references the default labels via GetServiceLabels, and updates it with the labels specific to the Service struct. This format was replicated for the NIM/NEMO resources that reference the Service struct
```
// GetExposeServiceLabels returns labels to apply to the generated Kubernetes Service.
func (n *NemoCustomizer) GetExposeServiceLabels() map[string]string {
	serviceLabels := n.GetServiceLabels()

	if n.Spec.Expose.Service.Labels != nil {
		return utils.MergeMaps(serviceLabels, n.Spec.Expose.Service.Labels)
	}
	return serviceLabels
}
...........
params.Labels = n.GetExposeServiceLabels()
```
TESTING:
Created _TestGetExposeServiceLabels_ which makes sure the new spec.expose.service.labels field is scoped only to the generated Kubernetes Service. It verifies that those labels do not appear in GetServiceLabels(), which is the shared helper used by other NIMService-owned resources like Deployments and HPAs. It also verifies that GetExposeServiceLabels() includes the normal standard labels, top-level spec.labels, and the new Service-specific labels. The other tests were also updated for the new labels field.


visheshtanksale (approved):
LGTM!!

PR #866, opened 2026-06-22
Add priorityClass to NIMService

JunAr7112:
This MR is in response to this [issue](https://github.com/NVIDIA/k8s-nim-operator/issues/731). Essentially, we are trying to ensure the NIMService pods have priority for scheduling by using priority classes.

SOLUTION:

The NIMServiceSpec struct was updated to include a PriorityClassName field. This is passed to GetDeploymentParams(), GetLWSParams(), GetStatefulSetParams(), for the default path and GetInferenceServiceParams() for kserve.

type NIMServiceSpec struct {
	...........
	RuntimeClassName string `json:"runtimeClassName,omitempty"`
	// PriorityClassName is the name of the PriorityClass to assign to pods created by this NIMService.
	// If set, the Kubernetes scheduler uses the specified priority for preemption decisions,
	// ensuring pods are scheduled even when resources (e.g. GPUs) are fully utilised by lower-priority workloads.
	// The referenced PriorityClass must exist in the cluster.
	PriorityClassName string                     `json:"priorityClassName,omitempty"`
	Proxy             *ProxySpec                 `json:"proxy,omitempty"`

func (n *NIMService) GetDeploymentParams() *rendertypes.DeploymentParams {
	............
        params.PriorityClassName = n.GetPriorityClassName()
        

TESTING:

TestGetPriorityClassName builds a minimal NIMService with spec.priorityClassName: gpu-critical and checks that the value is read correctly and passed into both GetDeploymentParams() (standalone) and GetInferenceServiceParams() (KServe). This makes sure the CR field exists and is wired through to the templates.

### DIFF
--- a/api/apps/v1alpha1/nimservice_types.go
+++ b/api/apps/v1alpha1/nimservice_types.go
@@ -125,12 +125,15 @@ type NIMServiceSpec struct {
 	SchedulerName  string        `json:"schedulerName,omitempty"`
 	Metrics        Metrics       `json:"metrics,omitempty"`
 	// +kubebuilder:validation:Minimum=0
-	Replicas         *int32                     `json:"replicas,omitempty"`
-	UserID           *int64                     `json:"userID,omitempty"`
-	GroupID          *int64                     `json:"groupID,omitempty"`
-	RuntimeClassName string                     `json:"runtimeClassName,omitempty"`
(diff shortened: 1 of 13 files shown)


visheshtanksale (approved):
LGTM!!

Issue #829, opened 2026-04-22
NIMService: Knative annotations (autoscaling.knative.dev/*, serving.knative.dev/progress-deadline) not propagated to Revision in KServe serverless mode


oleksii-donets:
### 1. Quick Debug Information
* OS/Version:
* Kernel Version:
* Container Runtime Type/Version:
* K8s Flavor/Version:
* GPU Operator Version:
* NIM Operator Version: 3.1.0
* LLM NIM Versions: nvcr.io/nim/meta/llama-3.2-1b-instruct (NIM LLM 1.12.0)
* NeMo Service Versions: N/A

### 2. Issue or feature description

When using `inferencePlatform: kserve` with KServe's Knative (serverless) deployment mode, Knative-specific annotations placed on `NIMService.metadata.annotations` are not propagated end-to-end to the underlying Knative `Revision`.

Annotations set on the `NIMService`:

- `autoscaling.knative.dev/initial-scale: "1"`
- `autoscaling.knative.dev/min-scale: "1"`
- `autoscaling.knative.dev/max-scale: "1"`
- `serving.knative.dev/progress-deadline: 3640s`

Observed on the resulting Knative `Revision` (inspected via `kubectl get revision <name> -o yaml`): only `autoscaling.knative.dev/min-scale` is present. `initial-scale`, `max-scale`, and `progress-deadline` are missing.

**Expected behavior:** Knative autoscaling annotations and `serving.knative.dev/progress-deadline` set on a `NIMService` should be forwarded to the `InferenceService` the operator creates so that KServe mirrors them onto the Knative `Configuration` / `Revision`.

**Actual behavior:** Only `min-scale` reaches the Revision. `max-scale`, `initial-scale`, and `progress-deadline` are silently dropped.

**Impact:** The dropped `progress-deadline` is the most damaging. Knative falls back to its default of 600s. NIM LLM containers frequently need more than 10 minutes on a first start (model download + load). As a result the Revision is aborted with `reason: ProgressDeadlineExceeded` / "ReplicaSet has timed out progressing" even when `startupProbe` would eventually succeed, and the `NIMService` is stuck in `NotReady` with `Configuration "…-predictor" is waiting for a Revision to become ready.`

### 3. Steps to reproduce the issue

1. Install NIM Operator 3.1.0 in a cluster with KServe + Knative Serving.
2. Apply a `NIMService` with `inferencePlatform: kserve` and the annotations below:

   ```yaml
[...]

PR #869, opened 2026-06-25
Propagate knative annotations in KServe serverless mode

JunAr7112:
This PR is in response to this [issue](https://github.com/NVIDIA/k8s-nim-operator/issues/829). Essentially, when deployign with KServe's Knative (serverless) deployment mode, Knative-specific annotations placed on NIMService.metadata.annotations are not propagated end-to-end to the underlying Knative Revision.

SOLUTION:
The code change adds GetInferenceServiceAnnotations() in nimservice_types.go. It starts with the existing service annotations, then selectively copies relevant top-level metadata.annotations onto the generated KServe resources. 

This is propagated in GetInferenceServiceParams():
params.Annotations = n.GetInferenceServiceAnnotations()
params.PodAnnotations = n.GetInferenceServiceAnnotations()

The controller update in nimservice_controller.go was also updated to reconcile for metadata.annotations.

TESTING:
Added a regression test in nimservice_types_test.go. It verifies that Knative annotations from metadata.annotations are propagated, unrelated annotations are ignored, and spec.annotations still wins if the same key exists.

### DIFF
--- a/api/apps/v1alpha1/nimservice_types.go
+++ b/api/apps/v1alpha1/nimservice_types.go
@@ -538,7 +538,26 @@ func (n *NIMService) GetNIMServiceAnnotations() map[string]string {
 	return standardAnnotations
 }
 
-// GetServiceLabels returns merged labels to apply to NIMService-owned resources.
+// GetInferenceServiceAnnotations returns annotations to apply to KServe InferenceService resources.
+func (n *NIMService) GetInferenceServiceAnnotations() map[string]string {
+	annotations := maps.Clone(n.GetNIMServiceAnnotations())
+	for key, value := range n.GetAnnotations() {
+		if shouldPropagateMetadataAnnotationToInferenceService(key) {
+			if _, exists := annotations[key]; !exists {
+				annotations[key] = value
+			}
+		}
+	}
+	return annotations
+}
+
+func shouldPropagateMetadataAnnotationToInferenceService(key string) bool {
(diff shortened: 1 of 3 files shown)


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -538,7 +538,26 @@ func (n *NIMService) GetNIMServiceAnnotations() map[string]string {
 	return standardAnnotations
 }
 
-// GetServiceLabels returns merged labels to apply to NIMService-owned resources.
+// GetInferenceServiceAnnotations returns annotations to apply to KServe InferenceService resources.
```
visheshtanksale, line 541: @JunAr7112 Did you test why the mentioned annotations are not passing on the inferenceservice object?

The `GetNIMServiceAnnotations` method seems to be already putting all the annotations mentioned on the NIMService on the InferenceService object, then why are the annotations mentioned in the issue on appearing on the inferenceservice.

JunAr7112: Right now, GetNimServiceAnnotations is only using the operator-generated annotations and
NIMService.spec.annotations, however, it is not taking in the NIMService.metadata.annotations. The user in the issue is adding annotations to NIMService.metadata which are not getting propagated. Since the NIM Operator is creating/managing a kserve Inference Service, the user expectation here is that optional knative annotations can be added to NIMService.metadata and would be propagated to the knative revision.

When I tested earlier, I verified metadata annotation were not included as seen in the bug

PR #870, opened 2026-06-28
Guard omitted NemoCustomizer training config map

fallintoplace:
## Summary
- guard the optional `trainingConfig.configMap` before reading its name in `addTrainingConfig`
- keep config rendering working when the training ConfigMap is omitted and rely on the generated training settings
- add a controller regression test that reconciles a `NemoCustomizer` with `trainingConfig.configMap` unset

## Why
`TrainingConfig.ConfigMap` is optional in the API, but the controller unconditionally dereferenced `n.Spec.Training.ConfigMap.Name` while rendering the customizer config. A valid CR that omitted the field could panic reconciliation.

## Validation
- `go test -ldflags "-X github.com/NVIDIA/k8s-dra-driver-gpu/internal/info.version=v25.12.0" ./internal/controller -run TestControllers -ginkgo.focus='NemoCustomizer Controller' -count=1`

### DIFF
--- a/internal/controller/nemocustomizer_controller.go
+++ b/internal/controller/nemocustomizer_controller.go
@@ -663,7 +663,8 @@ func (r *NemoCustomizerReconciler) addTrainingConfig(ctx context.Context, cfg ma
 	}
 
 	trainingCfg := make(map[string]interface{})
-	if cmName := n.Spec.Training.ConfigMap.Name; cmName != "" {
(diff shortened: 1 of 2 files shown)


visheshtanksale:
@fallintoplace NIM Operator is deprecating support for the NeMo Microservices CRDs and deployments in the next release. The NeMo Microservices are transitioning into the NeMo Platform. Refer to the [NeMo Platform documentation](https://docs.nvidia.com/nemo-platform/documentation/home#what-you-can-do) for details on continuing to use NeMo Microservices.
Closing this PR. Please reopen if you have any concerns

PR #871, opened 2026-06-28
Require a single NemoGuardrail config store source

fallintoplace:
## Summary
- require exactly one `spec.configStore.configMap` or `spec.configStore.pvc` for `NemoGuardrail`
- validate the config store at reconcile time so invalid objects fail cleanly instead of panicking
- harden the volume helpers and add focused Guardrail API/controller tests for the invalid config-store cases

## Why
`NemoGuardrail` only rejected the case where both config sources were set. If neither was set, the API accepted the object and `GetVolumes()` fell through to the PVC path and dereferenced `n.Spec.ConfigStore.PVC.Name` while `PVC` was nil.

This change closes the contract gap in two places: the CRD now requires exactly one config source, and the controller still returns a clear error if it encounters an invalid object from a stale CRD or another programmatic caller.

## Validation
- `go test ./api/apps/v1alpha1 -run 'TestNemoGuardrail' -count=1`
- `go test -ldflags "-X github.com/NVIDIA/k8s-dra-driver-gpu/internal/info.version=v25.12.0" ./internal/controller -run 'TestReconcileNemoGuardrailRejectsInvalidConfigStore' -count=1`

### DIFF
--- a/api/apps/v1alpha1/nemo_guardrails_types.go
+++ b/api/apps/v1alpha1/nemo_guardrails_types.go
@@ -118,7 +118,7 @@ type NIMEndpoint struct {
 
 // GuardrailConfig defines the source where the service config is made available.
 //
-// +kubebuilder:validation:XValidation:rule="!(has(self.configMap) && has(self.pvc))", message="Cannot set both ConfigMap and PVC in ConfigStore"
+// +kubebuilder:validation:XValidation:rule="has(self.configMap) != has(self.pvc)", message="Exactly one of configMap or pvc must be set in ConfigStore"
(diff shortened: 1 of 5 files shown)


visheshtanksale:
@fallintoplace NIM Operator is deprecating support for the NeMo Microservices CRDs and deployments in the next release. The NeMo Microservices are transitioning into the NeMo Platform. Refer to the [NeMo Platform documentation](https://docs.nvidia.com/nemo-platform/documentation/home#what-you-can-do) for details on continuing to use NeMo Microservices.
Closing this PR. Please reopen if you have any concerns

PR #872, opened 2026-06-28
Require DRA attribute selector values

fallintoplace:
## Summary
- require `attributeSelectors[].value` for shared DRA selector types used by NIMService and NIMPipeline CRDs
- harden DRA selector evaluation and validation paths so missing values return clear errors instead of dereferencing nil
- add focused API, webhook, and standalone controller regressions for omitted attribute selector values

## Why
`DRADeviceAttributeSelector.Value` was optional in the generated schema, but multiple code paths treated it as mandatory:
- `GetCELExpression()` dereferenced `d.Value`
- standalone DRA validation dereferenced `attribute.Value.VersionValue`
- webhook validation assumed the pointer itself was non-nil

That let invalid objects through the schema and turned them into panics or nil-pointer crashes during validation or reconciliation.

## Validation
- `go test ./api/apps/v1alpha1 -run TestDRADeviceAttributeSelectorGetCELExpression -count=1`
- `go test ./internal/webhook/apps/v1alpha1 -run TestValidateDRAResourcesConfiguration -count=1`
- `go test -ldflags "-X github.com/NVIDIA/k8s-dra-driver-gpu/internal/info.version=v25.12.0" ./internal/controller/platform/standalone -run TestControllers -ginkgo.focus='AttributeSelectors value is omitted' -count=1`

### DIFF
--- a/api/apps/v1alpha1/dra_types.go
+++ b/api/apps/v1alpha1/dra_types.go
@@ -87,6 +87,10 @@ type DRADeviceAttributeSelectorValue struct {
 }
 
 func (d *DRADeviceAttributeSelectorValue) GetValue() any {
+	if d == nil {
+		return nil
+	}
+
 	switch {
 	case d.BoolValue != nil:
 		return *d.BoolValue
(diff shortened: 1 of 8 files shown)


visheshtanksale (commented):
@fallintoplace This looks good. Can you please add sign off to the commit so that it can pass DCO check.
Also, initiated GH pipelines. Waiting on the pipelines to complete before approval

visheshtanksale (approved):

PR #873, opened 2026-06-30
Update NimService for CoCo

JunAr7112:
This MR has two commits:

The first commit updates the NIMService CRD for CoCo. We will Add the ConfidentialContainersSpec, and related InitDataSpec and ConfidentialModelSpec which will be used for the CoCo workflow.

The second commit will generate spec for init data for dry run. In nimservice_types.go we will create three functions:
IsConfidentialContainersEnabled() -> checks if the confidential containers flag is enabled
IsConfidentialDeploymentDryRun() -> checks if the dry run flag is enabled
SetConfidentialDeployment(deployment *appsv1.Deployment) -> creates a deepcopy of the deployment spec if the confidential containers flag is set

...in nimservice.go, renderAndSyncResource was updated to generate the Deployment spec for dry run and follow the normal spec skip behavior for other resources.

```
if isConfidentialDeploymentDryRun {
		if deployment, ok := resource.(*appsv1.Deployment); ok &&
			deployment.GetName() == nimService.GetName() &&
			deployment.GetNamespace() == nimService.GetNamespace() {
			nimService.SetConfidentialDeployment(deployment)
			return nil
		}
	}
```
TESTING: 
TestIsConfidentialDeploymentDryRun builds small NIMService objects with different confidentialContainers settings and verifies what IsConfidentialDeploymentDryRun() returns. The important cases are: disabled confidential containers should never count as DryRun, an empty phase should default to DryRun, explicit DryRun should be true, and explicit Active should be false.
TestSetConfidentialDeployment checks the in-memory Deployment setter. It first creates a representative full Deployment, including metadata, labels, ownerReferences, replicas, and runtime class. Then it verifies three things: the setter does nothing when confidential containers are disabled, stores the full Deployment when enabled, and stores it as a deep copy.


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -139,6 +148,151 @@ type NIMServiceSpec struct {
...
+	// Deployment provides an in-memory copy of the rendered Deployment to be verified with GenPolicy.
+	Deployment *appsv1.Deployment `json:"-"`
+
+	// SealedAuthSecret contains sealed secret reference.
+	SealedAuthSecret string `json:"sealedAuthSecret"`
+
+	// Model configures model delivery without NIMCache.
+	Model ConfidentialModelSpec `json:"model"`
+}
+
```
mkhaas, line 188: The current struct makes initData, sealedAuthSecret, and model required whenever confidentialContainers is present, even if enabled is false. Is that intentional? If these fields should only be required when confidential containers are enabled, we may need to make them optional and rely on conditional validation for the enabled: true case.

Additionally, there will be no initData for OpenShift

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -139,6 +148,151 @@ type NIMServiceSpec struct {
...
+//
+// +kubebuilder:validation:XValidation:rule="(has(self.ngc) ? 1 : 0) + (has(self.hf) ? 1 : 0) == 1",message="exactly one of ngc or hf must be set"
+type ConfidentialModelSpec struct {
+	// NGC configures pull from NVIDIA NGC.
+	// +optional
+	NGC *ConfidentialNGCModelSpec `json:"ngc,omitempty"`
+
+	// HF configures pull from Hugging Face Hub.
+	// +optional
+	HF *ConfidentialHFModelSpec `json:"hf,omitempty"`
```
mkhaas, line 255: We also support S3. (@visheshtanksale )

api/apps/v1alpha1/nimservice_types.go
JunAr7112: I can make initData optional for the OpenShift case. 

I was thinking the Enabled bool would be used if the user wants to temporarily disable CoCo verification and would default to True. Since the parent block is already optional there wouldn't be much need for a partially configured Confidential Containers struct.

internal/controller/platform/standalone/nimservice.go
```diff
@@ -942,14 +942,24 @@ func (r *NIMServiceReconciler) renderAndSyncResource(ctx context.Context, nimSer
...
+	// If confidential deployment dry run is enabled and the resource is a deployment, set the confidential deployment
+	if nimService.IsConfidentialDeploymentDryRun() {
+		if deployment, ok := resource.(*appsv1.Deployment); ok &&
+			deployment.GetName() == nimService.GetName() &&
+			deployment.GetNamespace() == nimService.GetNamespace() {
+			nimService.SetConfidentialDeployment(deployment)
+			return nil
+		}
+	}
+
```
mkhaas, line 954: Should we add validation that an Active deployment cannot move back to dry-run

mkhaas (commented):
A general comment -- Will NIMService CR get automatically deleted after dry-run?

api/apps/v1alpha1/nimservice_types.go
jojimt: Actually, making it optional will break coco. Do we have a flag for openshift? If so, let's leverage it to ignore initdata. We won't be using the nvidia runtimeclass in that case anyway, correct? Openshift should have its own coco runtimeclass.

Additionally, we also want to support a user specified init data.

 ```
   // Provided: KBS admin supplies the complete init-data document verbatim.
    // The operator does NOT run GenPolicy and does NOT synthesize aa.toml/cdh.toml.
    // +optional
    Provided *ProvidedInitDataSpec `json:"provided,omitempty"`

// ProvidedInitDataSpec supplies a complete init-data document.
// The document must be plain TOML (envelope: version, algorithm, [data]).
// It is NOT pre-encoded — the operator applies Encoding itself so the
// artifact ConfigMap stays human-reviewable and the digest is reproducible.
//
// +kubebuilder:validation:XValidation:rule="has(self.inline) != has(self.configMapRef)",message="exactly one of inline or configMapRef must be set"
type ProvidedInitDataSpec struct {
    // Inline is the full init-data TOML.
    // +kubebuilder:validation:MaxLength=131072
    // +optional
    Inline *string `json:"inline,omitempty"`
    // ConfigMapRef references a KBS-admin ConfigMap holding the init-data TOML.
    // +optional
    ConfigMapRef *ConfigMapKeyRef `json:"configMapRef,omitempty"`
}
```

visheshtanksale:
Not part of the current release plan. Closing this.

PR #886, opened 2026-07-21
feat: support native NIM model downloads

a-rich:

PR #889, opened 2026-07-23
fix: compress large model manifests to fit ConfigMap size limit

anhtra3889:
## Summary

`NIMCache` (and `NIMBuild`) store the extracted `model_manifest.yaml` — and the
built `local_model_manifest.yaml` — in a `ConfigMap`'s `Data` field. For models
with many profiles the manifest exceeds the hard **1 MiB (1048576 bytes)**
Kubernetes ConfigMap limit, so reconciliation fails and the model can never be
cached.

Example: `nemotron-3-ultra` produces a **~1.63 MiB** manifest, and the operator
fails with:

```
ConfigMap "<name>" is invalid: []: Too long: may not be more than 1048576 bytes
```

This affects large models regardless of GPU type (H100/H200/B200/B300) and
regardless of the profile requested — the operator stores the *full* manifest
before any profile filtering.

## Fix

Introduce a **threshold-hybrid** storage scheme in `internal/utils`:

- `SetManifestConfigMapData(cm, key, raw)`
  - `len(raw) <= 900 KiB` → store as plaintext in `Data[key]` (unchanged behavior)
  - otherwise → gzip-compress into `BinaryData[key + ".gz"]`
  - removes the counterpart representation so only one form is ever present
- `GetManifestConfigMapData(cm, key)`
  - reads transparently, preferring the compressed form, falling back to plaintext

These helpers are wired into:

- `nimcache_controller.go`: `createManifestConfigMap` (write) and `extractNIMManifest` (read)
- `nimbuild_controller.go`: `updateManifestConfigMap` (write of `local_model_manifest.yaml`)

The `nemotron-3-ultra` manifest compresses from **~1.63 MiB → ~0.34 MiB**, well
under the limit (~4.8× ratio).

## Backward compatibility

- Small manifests keep the exact plaintext form, so behavior is unchanged for the common case.
- Readers handle **both** representations, so existing ConfigMaps continue to work with **no migration**.
- The manifest ConfigMap is only read in-process by the operator (never mounted as a volume), so moving large payloads to `BinaryData` has no consumer-side impact.
- Minor downgrade caveat: a manifest written by a new operator in compressed form would not be readable by an *older* operator that only reads `Data[...]`. Small (plaintext) manifests remain fully downgrade-safe.

## Testing

- Unit tests (`internal/utils`): gzip round-trip (incl. empty + corrupt input), plaintext/gzip threshold (below / at / above), stale-entry cleanup on switch, read precedence/fallback/not-found/corrupt, and a `Set`→`Get` round-trip table across sizes.
- Controller test (`internal/controller`): `extractNIMManifest` transparently decompresses and parses a gzip-stored manifest ConfigMap.
- Verified end-to-end against a **real kube-apiserver** (envtest, which enforces the real 1 MiB limit) using the actual 1.63 MiB `nemotron-3-ultra` manifest:
  - plaintext `ConfigMap.Create` → rejected with `Too long` (reproduces the bug)
  - compressed `ConfigMap.Create` → succeeds
  - round-trip read is byte-identical to the original manifest

```
go build ./...                     # ok
go vet ./internal/...              # ok
go test ./internal/utils/...       # ok
go test ./internal/controller/...  # ok
```

## Notes / relationship to #886

This is orthogonal to #886 (native model download protocol). #886 only bypasses
the manifest-ConfigMap path for images labeled
`com.nvidia.nim.model_download_protocol=native-v1`. Standard LLM images that use
the legacy manifest/profile flow (e.g. `nemotron-3-ultra`, which does **not**
carry that label) still hit the 1 MiB limit; this PR fixes that path directly,
and the two changes can coexist.

## How to reproduce (before this fix)

```yaml
kind: NIMCache
metadata:
  name: nemotron-3-ultra
spec:
  source:
    ngc:
      modelPuller: nvcr.io/nim/nvidia/nemotron-3-ultra-ga:<tag>
      pullSecret: ngc-secret
  storage:
    pvc:
      create: true
      storageClass: <sc>
      size: "500Gi"
```

The `NIMCache` goes to `Failed` with the `Too long: may not be more than 1048576 bytes` error in the operator logs.

### DIFF
--- a/internal/controller/nimbuild_controller.go
+++ b/internal/controller/nimbuild_controller.go
@@ -1039,22 +1039,28 @@ func (r *NIMBuildReconciler) updateManifestConfigMap(ctx context.Context, nimCac
 		},
 	}
 
-	// Create the ConfigMap
+	// Create the ConfigMap. Small manifests are stored as plaintext for backward
+	// compatibility; large ones are gzip-compressed into binaryData so the
+	// ConfigMap stays under the 1 MiB limit (mirrors NIMCache manifest storage).
+	var setErr error
 	if err := k8sutil.RetryUpdate(ctx, r.Client, configMap, func(obj client.Object) {
 		cm, ok := obj.(*corev1.ConfigMap)
 		if !ok {
 			return
 		}
-		if cm.Data == nil {
-			cm.Data = make(map[string]string)
+		if _, err := utils.SetManifestConfigMapData(cm, LocalModelManifestKey, prettyManifestBytes); err != nil {
+			setErr = err
+			return
 		}
-		cm.Data["local_model_manifest.yaml"] = string(prettyManifestBytes)
 		if cm.Labels == nil {
 			cm.Labels = make(map[string]string)
 		}
 		cm.Labels["app"] = nimCache.GetName()
 	}); err != nil {
 		return fmt.Errorf("failed to create manifest ConfigMap %s: %w", configMap.Name, err)
+	}
+	if setErr != nil {
+		return fmt.Errorf("failed to set local manifest data for ConfigMap %s: %w", configMap.Name, setErr)
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


JunAr7112:
Thanks for the PR @anhtra3889 will review this fix

visheshtanksale:
closing in favor for https://github.com/NVIDIA/k8s-nim-operator/pull/891

Issue #890, opened 2026-07-23
NIMService (KServe serverless): scale-to-zero impossible — min-scale/max-scale annotations are stripped by KServe; operator should map scale bounds to predictor minReplicas/maxReplicas


oleksii-donets:
### 1. Quick Debug Information
* OS/Version:
* Kernel Version:
* Container Runtime Type/Version:
* K8s Flavor/Version: K8s v1.34.1
* GPU Operator Version:
* NIM Operator Version: 3.1.0 (behavior confirmed against current `main` sources as well)
* KServe Version: v0.18.0 (serverless / Knative deployment mode)
* LLM NIM Versions: nvcr.io/nim/meta/llama-3.2-1b-instruct
* NeMo Service Versions: N/A

### 2. Issue or feature description

Follow-up to #829. As recommended there, Knative autoscaling annotations are now supplied via `NIMService.spec.annotations`, and the operator propagates them to the `InferenceService` correctly. However, in KServe **serverless (Knative)** deployment mode, `autoscaling.knative.dev/min-scale` and `autoscaling.knative.dev/max-scale` still never take effect, because KServe itself strips exactly these two keys and regenerates them from the predictor's typed fields.

**Root cause (two halves):**

1. KServe explicitly disallows pass-through of these two annotations — `ServiceAnnotationDisallowedList` in [`pkg/constants/constants.go`](https://github.com/kserve/kserve/blob/master/pkg/constants/constants.go) contains `autoscaling.knative.dev/min-scale` and `autoscaling.knative.dev/max-scale`. KServe then sets `min-scale` from `predictor.minReplicas` (defaulting to **1** when nil) and `max-scale` from `predictor.maxReplicas` (omitted when unset). In KServe serverless mode these typed fields are the *only* supported way to control scale bounds — and `minReplicas: 0` is the only way to enable scale-to-zero.

[...]

PR #897, opened 2026-07-30
Update Kserve Serverless Deployment for scale-to-zero

JunAr7112:
This PR is a response to this [issue](https://github.com/NVIDIA/k8s-nim-operator/issues/890), and is aimed at ensuring that in Kserve serverless mode the, the operator can scale replicas to zero.

**api/apps/v1alpha1/nimservice_types.go** - Created a helper function _parseKnativeScaleAnnotation(...)_ to turn a Knative annotation (min-scale/max-scale) into an *int32. It returns (nil, nil) when the annotation is absent or empty, and an error when the value is not a valid non-negative int32.
...In KServe serverless (Knative) deployment mode, autoscaling.knative.dev/min-scale[or max-scale] still never take effect, because KServe strips these two keys and regenerates them from the predictor's typed fields. As such, in serverless deployment mode we will include:
```
if minReplicas, err := parseKnativeScaleAnnotation(params.Annotations, knativeMinScaleAnnotation); err == nil && minReplicas != nil {
			params.MinReplicas = minReplicas
		}
		if maxReplicas, err := parseKnativeScaleAnnotation(params.Annotations, knativeMaxScaleAnnotation); err == nil && maxReplicas != nil && *maxReplicas > 0 {
			params.MaxReplicas = maxReplicas
		}
```

**manifests/inferenceservice.yaml** - Setting params.MaxReplicas in Go isn't enough on its own, because the template only rendered maxReplicas for raw/standard. We moved it out of that gate so it renders in serverless too:
```
{{- if .MinReplicas }}
minReplicas: {{ .MinReplicas }}
{{- end }}
{{- if .MaxReplicas }}
maxReplicas: {{ .MaxReplicas }}
{{- end }}
{{- if or (eq .DeploymentMode "RawDeployment") (eq .DeploymentMode "Standard") }}
```
**TESTING**
_TestParseKnativeScaleAnnotation_ unit-tests the _parseKnativeScaleAnnotation_ helper across all input classes (absent, empty, "0", positive, negative, non-numeric), with the key assertion being that "0" returns a non-nil pointer to 0 — distinct from nil — since that nil-vs-zero distinction is exactly what makes scale-to-zero expressible. _TestGetInferenceServiceParamsKnativeScale_ then checks the Knative branch of GetInferenceServiceParams: given min-scale/max-scale annotations it sets params.MinReplicas/MaxReplicas correctly (scale-to-zero with a max bound), treats max-scale: "0" as unbounded (leaves max nil), leaves both untouched when no annotations are present, ignores malformed values, and never leaks the HPA autoscaler-class annotation into serverless mode.

### DIFF
--- a/api/apps/v1alpha1/nimservice_types.go
+++ b/api/apps/v1alpha1/nimservice_types.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"maps"
 	"os"
+	"strconv"
 	"strings"
 
 	kserveconstants "github.com/kserve/kserve/pkg/constants"
@@ -1840,6 +1841,34 @@ func (n *NIMService) GetHostPath() string {
 	return ""
 }
 
+const (
+	// knativeMinScaleAnnotation is the Knative annotation for the minimum number
+	// of pods (autoscaling.knative.dev/min-scale). A value of "0" enables
+	// scale-to-zero.
+	knativeMinScaleAnnotation = "autoscaling.knative.dev/min-scale"
+	// knativeMaxScaleAnnotation is the Knative annotation for the maximum number
+	// of pods (autoscaling.knative.dev/max-scale). A value of "0" means unbounded.
+	knativeMaxScaleAnnotation = "autoscaling.knative.dev/max-scale"
+)
+
+// parseKnativeScaleAnnotation parses a Knative scale annotation (min-scale /
+// max-scale) into an *int32. It returns (nil, nil) when the annotation is absent
+// or empty, and an error when the value is not a valid non-negative int32.
+func parseKnativeScaleAnnotation(annotations map[string]string, key string) (*int32, error) {
+	value, ok := annotations[key]
+	if !ok || value == "" {
+		return nil, nil
+	}
+	parsed, err := strconv.ParseInt(value, 10, 32)
+	if err != nil {
+		return nil, fmt.Errorf("invalid %s annotation %q: %w", key, value, err)
+	}
+	if parsed < 0 {
+		return nil, fmt.Errorf("invalid %s annotation %q: must not be negative", key, value)
+	}
+	return ptr.To(int32(parsed)), nil
+}
+
 // GetInferenceServiceParams returns params to render InferenceService from templates.
 func (n *NIMService) GetInferenceServiceParams(
 	deploymentMode kserveconstants.DeploymentModeType) *rendertypes.InferenceServiceParams {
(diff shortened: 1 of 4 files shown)


api/apps/v1alpha1/nimservice_types.go
```diff
@@ -1886,6 +1915,24 @@ func (n *NIMService) GetInferenceServiceParams(
...
+		// In KServe serverless (Knative) mode, scaling is handled by the Knative
+		// KPA. KServe keeps autoscaling.knative.dev/min-scale and max-scale on its
+		// ServiceAnnotationDisallowedList and regenerates them from the typed
+		// predictor.minReplicas/maxReplicas fields (min-scale defaults to 1 when
+		// minReplicas is nil, which makes scale-to-zero impossible). HPA-based
+		// autoscaling (spec.scale.enabled) is rejected by the webhook in this mode,
+		// so users express scale bounds via the Knative annotations. Translate those
+		// annotations into the typed fields here so they survive KServe rendering.
+		// A minReplicas of 0 enables scale-to-zero.
+		if minReplicas, err := parseKnativeScaleAnnotation(params.Annotations, knativeMinScaleAnnotation); err == nil && minReplicas != nil {
```
visheshtanksale: This is failing silently. Needs to atleast log the error message

api/apps/v1alpha1/nimservice_types.go
```diff
@@ -1886,6 +1915,24 @@ func (n *NIMService) GetInferenceServiceParams(
...
+			params.MinReplicas = minReplicas
+		}
+		// max-scale of 0 means "unbounded" in Knative; leave MaxReplicas unset so
+		// KServe omits max-scale rather than pinning it to 0.
+		if maxReplicas, err := parseKnativeScaleAnnotation(params.Annotations, knativeMaxScaleAnnotation); err == nil && maxReplicas != nil && *maxReplicas > 0 {
```
visheshtanksale: Same as above. Atleast needs a log message

api/apps/v1alpha1/nimservice_types.go
JunAr7112: Alright I logged the error message. In the case where parseKnativeScaleAnnotation fails, the error will be logged  and minReplicas will default to 1

api/apps/v1alpha1/nimservice_types.go
JunAr7112: Alright I logged the error message

visheshtanksale (approved):

PR #904, opened 2026-08-03
Update skills for uninstall and credentials

JunAr7112:
**.agents/skills/nim-operator-install/scripts/validate-nim-operator-install.sh** - Made the following changes:
Bound every kubectl call so an unreachable/black-hole API endpoint cannot stall the diagnostics for the full default TCP timeout. Override via KUBECTL_REQUEST_TIMEOUT (for example "30s") for slow or large clusters.
```
kubectl() {
  command kubectl --request-timeout="${KUBECTL_REQUEST_TIMEOUT:-10s}" "$@"
}
```
Report failures instead of aborting: this is a diagnostic script and must keep running (and covering later sections) even when an individual probe fails (for example an unreachable cluster, or "kubectl auth can-i" returning "no").
```
run() {
  printf '+ %s\n' "$*"
  "$@"
  "$@" || printf '  (command failed: exit %s)\n' "$?"
}
```

**.agents/skills/nim-operator-uninstall/SKILL.md**
Updated the uninstall documentation to ensure custom resources are deleted before the Helm release is uninstalled, allowing the operator to clean up owned resources and remove finalizers before it is removed. Added a verification step to confirm all custom resources have been fully deleted before proceeding, preventing race conditions that can leave resources stuck. Included warnings explaining the role of finalizers and the consequences of uninstalling the operator too early, and updated the introductory questions so they consistently reflect the safe uninstall order.

### DIFF
--- a/.agents/skills/nim-operator-install/scripts/validate-nim-operator-install.sh
+++ b/.agents/skills/nim-operator-install/scripts/validate-nim-operator-install.sh
@@ -1,17 +1,31 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+# Detach stdin so kubectl cannot block on an interactive credential prompt
+# (for example "Please enter Username:") when the kubeconfig lacks credentials.
+exec </dev/null
+
 namespace="${NIM_OPERATOR_NAMESPACE:-nim-operator}"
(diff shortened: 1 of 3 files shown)


visheshtanksale (approved):
LGTM

PR #907, opened 2026-08-04
Fix kserve with InferenceService and Standalone race conditions

JunAr7112:
**Issue 1**: NIM Operator starts before the InferenceService CRD exists and there is no isvc watch, when the isvc later becomes Ready, nothing runs NIMService reconcile so its status remains NotReady.

Solution:
Created two new constants - If a KServe NIMService is reconciled and the CRD still isn’t there, wait 30s and try again
const _inferenceServiceCRDRequeueInterval_ = 30 * time.Second
var _errInferenceServiceCRDMissing_ = errors.New("InferenceService CRD is not available")

For KServe NIMServices, ensure we are watching InferenceServices. If the CRD was missing at operator start-up, register the owner watch on first demand instead of polling discovery in the background.
```
	if nimService.Spec.InferencePlatform == appsv1alpha1.PlatformTypeKServe {
		if err := r.ensureInferenceServiceWatch(); err != nil {
			if errors.Is(err, errInferenceServiceCRDMissing) {
				logger.Info("InferenceService CRD not available yet; requeueing",
					"NIMService", nimService.Name)
				return ctrl.Result{RequeueAfter: inferenceServiceCRDRequeueInterval}, nil
			}
```
SetupWithManager: stop treating “no CRD at boot” as permanent
```
isvcCRDExists, err := k8sutil.CRDExists(...)
if isvcCRDExists {
    nimServiceBuilder = nimServiceBuilder.Owns(&kservev1beta1.InferenceService{})
    r.inferenceServiceWatchRegistered = true
}
```
**internal/controller/platform/kserve/nimservice.go** - for checkInferenceServiceStatus we will requeue on NotReady. Prefer the API reader so readiness is observed even when the InferenceService informer was not started (CRD missing at operator boot / not in filtered cache).
```
	reader := client.Reader(r.Client)
	if r.apiReader != nil {
		reader = r.apiReader
	}
```

**Issue 2**: when spec.scale.enabled is flipped to false while the HPA is actively rescaling, the HPA is deleted but the Deployment keeps the HPA's replica count forever, this breaks kubectl wait, GitOps health checks and any status-driven automation; symptom 2 also silently holds GPUs.

Solution:
Created two new constants:
_deploymentNotReadyRequeue_ -> wait 30s to re-check Deployment when the NIMService is not yet Ready.
_autoscalingDisableRequeue_ -> a short requeue used after disabling autoscaling so a late HPA write cannot permanently pin Deployment replicas

We will define a variable autoscalingJustDisabled which corresponds to the scenario where HPA was present and we just deleted it. Then we will proceed to delete the HPA as we did before.
```
if hpaGetErr == nil {
	autoscalingJustDisabled = true
}
```
Inside _renderAndSyncResource_, before if parent-spec hash matches we would return immediately (do nothing) leading to the replicas being stuck at 3. Now we skip only if parent-spec hash matches and live replicas already match spec.replicas.
```
parentSpecChanged := !found || utils.IsParentSpecChanged(obj, utils.DeepHashObject(nimService.Spec))
	replicasOutOfSync := false
	if found && !nimService.IsAutoScalingEnabled() {
		if curr, ok := obj.(*appsv1.Deployment); ok {
			replicasOutOfSync = deploymentSpecReplicasOutOfSync(curr, nimService.GetReplicas())
		}
	}
	if found && !parentSpecChanged && !replicasOutOfSync {
		return nil
	}
```
After disable you often are NotReady for a while (pods terminating). If a late HPA write left incorrect replicas you would be stuck. Now we will requeue in 30s.
```
if !ready {
	..................
	return ctrl.Result{RequeueAfter: deploymentNotReadyRequeue}, nil
}
```
Once the deployment is Ready, we add a check for where autoscalingJustDisabled=True and RequeueAfter 5s to catch late HPA writes. If a late HPA write set deployment back to 3, without a requeue, we might never look again (spec hash unchanged → skip). With the 5s requeue, we come back, see replicas 3 ≠ 1, force sync back to 1.

TESTING: For issue 1 - We verified issue 1 by reproducing the install-order bug on a live cluster: the fixed NIM Operator image was started first (with no InferenceService CRD present), then KServe v0.19 was installed afterward without restarting the operator. Creating a inferencePlatform: kserve, and once the owned InferenceService reported PredictorReady=True / Ready=True, the NIMService converged to Ready on the same operator pod that had booted before KServe—confirming status no longer stays stuck NotReady until a restart. 
For issue 2 we simulates the bug: HPA has scaled the Deployment to 3, then autoscaling is turned off with replicas: 1. It checks that reconcile deletes the HPA, forces Deployment replicas back to 1, and returns a non-zero RequeueAfter (not a silent no-op). It then fakes a late HPA write that bumps replicas to 3 again without changing the parent-spec hash, and asserts the next reconcile still restores replicas to 1.

### DIFF
--- a/internal/controller/nimservice_controller.go
+++ b/internal/controller/nimservice_controller.go
@@ -18,8 +18,11 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
+	"sync"
+	"time"
 
 	"github.com/go-logr/logr"
 	kservev1beta1 "github.com/kserve/kserve/pkg/apis/serving/v1beta1"
@@ -37,11 +40,13 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 	nvidiaresourcev1beta1 "sigs.k8s.io/dra-driver-nvidia-gpu/api/nvidia.com/resource/v1beta1"
 	lwsv1 "sigs.k8s.io/lws/api/leaderworkerset/v1"
 
@@ -59,6 +64,12 @@ import (
 // NIMServiceFinalizer is the finalizer annotation.
 const NIMServiceFinalizer = "finalizer.nimservice.apps.nvidia.com"
 
+// inferenceServiceCRDRequeueInterval is how long to wait before retrying when a
+// KServe NIMService is reconciled but the InferenceService CRD is not yet installed.
+const inferenceServiceCRDRequeueInterval = 30 * time.Second
+
+var errInferenceServiceCRDMissing = errors.New("InferenceService CRD is not available")
+
 // NIMServiceReconciler reconciles a NIMService object.
 type NIMServiceReconciler struct {
 	client.Client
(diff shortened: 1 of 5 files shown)


visheshtanksale (approved):
LGTM

Issue #887, opened 2026-07-22
Not able to create a NIMService without the authSecret field incase of disconnected environment and eventhough I have model cached to pvc it's still trying to authenticate to ngc.


SanthoshT2001:
### 1. Quick Debug Information
* OS/Version(e.g. RHEL8.6, Ubuntu22.04):
* Kernel Version:
* Container Runtime Type/Version(e.g. Containerd, CRI-O, Docker):
* K8s Flavor/Version(e.g. K8s, OCP, Rancher, GKE, EKS):
* GPU Operator Version: 26.3.x
* NIM Operator Version: 3.1.0 (https://docs.nvidia.com/nim-operator/3.1.0/service.html)
* LLM NIM Versions: 
* NeMo Service Versions:

### 2. Issue or feature description
_Briefly explain the issue in terms of expected behavior and current behavior._

1. Current behaviour: 
I have created a NIMService without setting authSecret field in disconnected environment since I have model cache in my server created pvc and passing to nim service.

Error: **The NIMService "nim-llm-service" is invalid:
* spec.authSecret: Required value
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation**

2. since the field is mandatory passed the key as:
apiVersion: apps.nvidia.com/v1alpha1
kind: NIMService
metadata:
  name: nim-llm-service
  namespace: nim-llm
spec:
  image:
    repository: nvcr.io/nim/meta/llama-3.2-1b-instruct
    tag: latest
  authSecret: ngc-api-secret
  env:
    - name: NIM_MODEL_PROFILE
      value: "72104c5b2318b83bb8dde1d255998243472158e48c236982c4974f5a1c693d8e"
    - name: NIM_CACHE_PATH
      value: "/opt/nim/.cache"
  storage:
    pvc:
      name: nim-cache-pvc
      mountPath: /opt/nim/.cache
  replicas: 1
  resources:
    limits:
      nvidia.com/gpu: 1

then it's trying authenticate to ngc catalog since we passed the authSecret:

<img width="3043" height="1722" alt="Image" src="https://github.com/user-attachments/assets/412530fa-7ba2-4f98-90b5-a4fcd97ecc9e" />

Expected behaviour: 
incase of disconnected environment when we have required model cache pvc in cluster, it should not check for authSecret/no authentication to ngc catalog.

### 3. Steps to reproduce the issue
_Detailed steps to reproduce the issue._

[...]

PR #910, opened 2026-08-07
Remove authSecret reqs for Air gapped use

JunAr7112:
This PR is a response to this issue: https://github.com/NVIDIA/k8s-nim-operator/issues/887. Essentially the NIMService can't be created without authSecret. If you include that secret, the operator automatically puts NGC_API_KEY into the pod which causes an error in disconnected environments.

SOLUTION:
**api/apps/v1alpha1/nimservice_types.go** - Make the AuthSecret optional to enable air-gapped deployments that serve a pre-cached model
```
type NIMServiceSpec struct {
	.........................
	AuthSecret string `json:"authSecret,omitempty"`
```
...Remove the unconditional block for NGCAPIKey, and switch to:
```
// Only inject NGC_API_KEY when AuthSecret is set...
if n.Spec.AuthSecret != "" {
	envVars = append(envVars, corev1.EnvVar{
		Name: NGCAPIKey,

```
**internal/webhook/apps/v1alpha1/nimservice_webhook_validation_helper.go** - Similarly, for the webhook we can allow empty for an AuthSecret and provide a warning here instead:
```
if authSecret == nil || *authSecret == "" {
		warningList = append(warningList, fmt.Sprintf("%s is empty: NGC_API_KEY will not be injected. Omit this only for air-gapped deployments with a pre-populated local model cache", fldPath.String()))
```
In Standalone + KServe controllers — don’t automatically re-add secret refs when empty.

TESTING: The new/updated tests check the two main behaviors of the fix. TestGetStandardEnvAuthSecret verifies that when authSecret is set, GetStandardEnv() still injects NGC_API_KEY from that secret, and when it is omitted, NGC_API_KEY is not added at all. TestValidateAuthSecret was updated so an empty authSecret is accepted with a warning instead of being rejected as a validation error. Also verified with an airgapped cluster:
```
local-agadiyar@ipp2-2404:~/k8s-nim-operator$ kubectl -n nim-llm get nimservice nim-llm-service -o jsonpath='authSecret={.spec.authSecret} status={.status.state}{"\n"}'
authSecret= status=Ready
local-agadiyar@ipp2-2404:~/k8s-nim-operator$ POD=$(kubectl -n nim-llm getpod -l app=nim-llm-service -o jsonpath='{.items[0].metadata.name}')
local-agadiyar@ipp2-2404:~/k8s-nim-operator$ kubectl -n nim-llm get pod "$POD" -o jsonpath='{range .spec.containers[0].env[*]}{.name}{"="}{.value}{"\n"}{end}' | grep -E 'NGC_API_KEY|NIM_CACHE_PATH|NIM_MODEL_PROFILE' || echo 'PASS: NGC_API_KEY absent'
NIM_MODEL_PROFILE=72104c5b2318b83bb8dde1d255998243472158e48c236982c4974f5a1c693d8e
NIM_CACHE_PATH=/model-store
local-agadiyar@ipp2-2404:~/k8s-nim-operator$ kubectl -n nim-llm get netpol bug-repro-deny-egress
NAME                    POD-SELECTOR          AGE
bug-repro-deny-egress   app=nim-llm-service   23h
local-agadiyar@ipp2-2404:~/k8s-nim-operator$ kubectl -n nim-llm exec "$POD" -- python3 -c '
import json,urllib.request
body=json.dumps({"model":"meta/llama-3.2-1b-instruct","messages":[{"role":"user","content":"Say hello in one short sentence."}],"max_tokens":16,"temperature":0}).encode()
req=urllib.request.Request("http://127.0.0.1:8000/v1/chat/completions",data=body,headers={"Content-Type":"application/json"})
print(json.load(urllib.request.urlopen(req,timeout=60))["choices"][0]["message"]["content"])
'
Hello.
```

### DIFF
--- a/api/apps/v1alpha1/nimservice_types.go
+++ b/api/apps/v1alpha1/nimservice_types.go
@@ -99,8 +99,9 @@ type NIMServiceSpec struct {
 	Command []string        `json:"command,omitempty"`
 	Args    []string        `json:"args,omitempty"`
 	Env     []corev1.EnvVar `json:"env,omitempty"`
-	// The name of an existing pull secret containing the NGC_API_KEY
-	AuthSecret string `json:"authSecret"`
+	// AuthSecret is the name of an existing secret containing the NGC_API_KEY (and optionally HF_TOKEN).
+	// It is optional for air-gapped deployments that serve a pre-cached model
+	AuthSecret string `json:"authSecret,omitempty"`
 	// Storage is the target storage for caching NIM model if NIMCache is not provided
 	Storage      NIMServiceStorage   `json:"storage,omitempty"`
 	Labels       map[string]string   `json:"labels,omitempty"`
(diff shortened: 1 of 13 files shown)


visheshtanksale (approved):
LGTM

PR #911, opened 2026-08-07
Add a THIRD_PARTY_NOTICES.md generator

abrarshivani:
Adds `THIRD_PARTY_NOTICES.md`, the license of every Go dep the operator image
ships plus the verbatim license text, and the script that generates it.

- `make third-party-notices` regenerates it
- `make check-third-party-notices` regenerates and diffs it, and runs in CI on every build
- Committed to the repository, not copied into the image, so image contents
  are unchanged

## What to review

98% of the diff is generated: `THIRD_PARTY_NOTICES.md`, 32,375 of 33,018 lines.
The other 682, the generator being the part worth reading:

| File | Lines |
|---|---:|
| `hack/generate-third-party-notices.sh` | 500 |
| `tools/go.sum` | 86 |
| `.github/workflows/golang.yml` | 31 |
| `Makefile` | 18 |
| `tools/go.mod` | 5 |
| `.gitignore` | 2 |
| `tools/tools.go` | 1 |

`tools/go.sum` is `go mod tidy` output.

## How it works

`make third-party-notices` runs `hack/generate-third-party-notices.sh`:

1. Resolve `go-licenses` from `./bin`, pinned in `tools/go.mod`, building it if
   missing.
2. Read `DOCKER_BUILD_PLATFORM_OPTIONS` from
   `deployments/container/multi-arch.mk` and fail if it is out of sync with the
   script's platform list, so a new release target cannot silently produce an
   incomplete file.
3. Read the clone URL, pinned commit and built path for the second image binary
   out of `deployments/container/Dockerfile`, failing on anything it cannot
   parse.
4. Run `go-licenses save` and `csv` over `./cmd/...` once per released platform,
   from `vendor/` with `CGO_ENABLED=0` to match the Dockerfile, then union.
   `go-licenses` resolves only the host platform and build-tagged sources differ
   per target, so one pass is incomplete and host-dependent.
5. Fetch the second binary's repo at the pinned commit into the scratch
   directory (never the worktree) and repeat there.
6. Collapse to one row per package, joining every distinct license.
7. Attach `module@version` by longest-prefix match against `vendor/modules.txt`.
   Where `go.mod` replaces a module the replacement is named, since that is what
   is vendored and shipped.
8. Emit the index tables, then each license file verbatim, each fence one
   backtick wider than the longest run in the file so licenses that are
   themselves Markdown do not break the document.

`make check-third-party-notices` regenerates and diffs, failing if the committed file is
stale or untracked. The `third-party-notices` job in `golang.yml` runs it on every build, not behind a
path filter. The list is the import set of the built packages, so it goes stale
when a `.go` file changes its imports, not only when `go.mod` or `vendor/` move.

## Implementation notes

- **Only the local module goes to `--ignore`.** It matches raw string prefixes,
  not path segments. The token `go` (from `go/ast`) would silently drop
  `golang.org/*`, `google.golang.org/*`, `gopkg.in/*` and `go.uber.org/*`.
- **Licenses are joined, not picked.** `sigs.k8s.io/yaml`'s single `LICENSE` file
  really does contain MIT, BSD-3-Clause and Apache text.
- **A guard rejects any unresolved `module@version`**, so a package that cannot
  be traced back to `vendor/modules.txt` fails the run.
- **Output must match on every machine**, since CI diffs it. `LC_ALL=C` on every
  sort and grep. Whole-line `sort -u`, because BSD sort keeps the first input row
  and GNU sort applies a whole-line tiebreak. The awk counts instead of testing
  `pkg in lic`, which mawk and busybox awk answer true on the first row,
  prefixing every license with `" / "`. `grep -a` so a NUL byte cannot change the
  fence width.
- **Moved into place with `mv`, not `cp`.** `cp` truncates the target first, so
  an interrupted run would leave a half-written file.
- **`go-licenses` needs the GOROOT the `go` command actually resolves** in each
  directory. It decides what is stdlib by comparing against the GOROOT compiled
  into its own binary, and this repo's `go.mod` names Go 1.26.5 while the cloned
  one names 1.26.0. Without the export every stdlib package looks module-less and
  the run dies with `Package os/signal does not have module info`.
- **A `git checkout` naming a branch or tag is rejected.** Only a full commit
  pins a dep set.

## Scope

Two shipped surfaces, each with its own index and license-text section.

**`manager`** is `./cmd/...`, exactly what the Dockerfile builds
(`go build ... cmd/main.go`). 168 packages from 156 modules. Against `./...` the
only extras are `github.com/onsi/ginkgo/v2` and its `golang.org/x/mod` and
`golang.org/x/tools` deps, reached only from the e2e suite under `test/`, which
is part of this module but never shipped.

**`crd-apply-tool`** is the second Go binary in the image, and not obvious from
this repo alone. The Dockerfile clones `NVIDIA/k8s-operator-libs`,
checks out `8e59492a772c39f8e99eea9cfc42567fcc1d939a`, builds
`examples/apply-crds/main.go` and copies it to `/usr/local/bin/crd-apply-tool`.
That module is not vendored here, and at the pinned commit resolves
`k8s.io/{api,client-go,apimachinery}` to v0.36.0 and `controller-runtime` to
v0.24.0, where the operator resolves them to v0.34.5 and v0.22.4 through replace
directives. The image ships both sets, so it gets its own index and license text:
merged into the runtime table, one row per package would have to discard a
version. 59 packages from 53 modules.

The build toolchain is deliberately not covered. `tools/go.mod` pins
`client-gen`, `lister-gen`, `informer-gen`, `controller-gen` and `kustomize`;
none appears in any Dockerfile, so none is redistributed. Attribution attaches to
what ships, so listing the generators behind committed artifacts overstated the
scope. That removed 65 of the previous 292 index rows.

Not covered, and the document header says so:

- The non-Go contents of the image, including everything inherited from the
  `nvcr.io/nvidia/distroless/go` base image, which goes through that image's own
  compliance process.
- `deployments/devel/go.mod`, the second tools module. It pins `golangci-lint`
  and `setup-envtest`, neither shipped nor producing a committed artifact. It
  also pins `controller-tools` at a different version from `tools/go.mod`
  (`v0.21.0` vs `v0.16.2`), and the
  `controller-gen`/`kustomize` targets read versions from there while
  `install-tools` builds from `tools/`. Predates this change, not resolved here.

Licenses observed across both surfaces: Apache-2.0, MIT, BSD-3-Clause, ISC,
BSD-2-Clause. Nothing outside the set NVIDIA already ships elsewhere.

## Shipping

Committed to the repository. Not copied into the image, so image size is
unchanged, and not attached to releases, so the release flow is untouched. An
earlier revision added `release-third-party-notices.yml`, which uploaded the file
as a release asset; that has been removed.

The check runs as the `third-party-notices` job in `golang.yml`, inheriting that
workflow's triggers rather than declaring its own.

Two consequences of covering `crd-apply-tool`:

- **The check now clones in CI**, so the gate needs `github.com` reachable.
  It is one pinned commit at `--depth 1`, and the job already downloads modules
  over the network.
- **Bumping the pinned commit means regenerating the notices in the same PR.**

## Testing

- **Cross-host.** Generated on macOS (BSD awk) and in `golang:1.26.5`
  (`/usr/bin/awk` is mawk). Identical sha256:
  `1c0700c9c9faea27ad23870e650bb983822b3de259b64b2fb8692c69f6c689d2`. Two
  consecutive runs byte-identical.
- **Completeness** against `go list -deps` for both released platforms, derived
  separately and compared with the index.
  - Runtime: 156 modules expected, 156 present, 0 missing, 0 extra; all 1,354
    non-stdlib packages covered by an index row.
  - `crd-apply-tool`: ground truth taken inside the cloned tree. 53 modules
    expected, 53 present, 0 missing; all 575 packages covered, 0 index rows
    outside the set.
- **Content.** 227 index rows across the two tables, every one with a license
  text section. 0 `Unknown` licenses, 0 "License text unavailable", 0 unresolved
  `module@version`, 0 URLs naming `HEAD`.
- **`CGO_ENABLED`.** Set to 0 because the Dockerfile builds `manager` that way.
  Checked for both Go surfaces: the module set is identical with cgo on and off.
  With the wrong value, build constraints can drop whole packages and
  `go-licenses` reports fewer rows, or none, while exiting 0.
- **`make check-third-party-notices`** passes on the committed file, fails when it is stale,
  and fails when it is untracked.
- **Error paths**, each failing with a specific message and a non-zero exit
  rather than writing a degraded file: missing `multi-arch.mk`; platform matrix
  out of sync; unparseable `DOCKER_BUILD_PLATFORM_OPTIONS`; unreadable
  `vendor/modules.txt`; unsafe `LICENSES_DIR`
  override; filesystem `replace` directive; missing Dockerfile; Dockerfile with
  no `git clone`, `git checkout` or `go build -o` line; `git checkout` naming a
  branch; pinned commit that cannot be fetched.
- **Lint.** `shellcheck` clean. On `golang.yml`, `actionlint` and
  `yamllint` report only the `linux-amd64-cpu4` unknown-runner-label warning and
  the missing-document-start and truthy-`on` warnings, all of which fire on every
  existing workflow here.


visheshtanksale (approved):
LGTM

PR #913, opened 2026-08-10
Fix legacy v1 parser and nimcache controller path for MolMIM

JunAr7112:
**Issue 1**: MolMIM 1.0.0 ships a legacy model_manifest.yaml (shown below). Right now, the operator only maps v1: hash → profile map and v2: schema_version + profiles: []. So NIMCache failed while parsing the manifest.
```
model_profile: MolMIM
manifest:
  - id: <hash>
    tags: {...}
```
**internal/nimparser/legacy.go** - Created a function _NormalizeLegacyManifest_, which detects if the data is of the form of the wrapped legacy shape and rewrites it to a v1 map. Returns (normalizedYAML, true, nil) when the input is legacy and conversion succeeds, returns (nil, false, nil) when the input is not a legacy wrapped manifest, returns a non-nil error when the input looks legacy but cannot be normalized.

**internal/nimparser/v1/nimparser.go** - Update _ParseModelManifest_ to call ParseModelManifestFromRawOutput which will call NormalizeLegacyManifest. If it is legacy it will follow the new pathway to create a normalized YAML. Otherwise it will use the existing v1 hash map directly. So any caller of the v1 parser gets MolMIM support automatically.

**internal/nimparser/utils/utils.go** - updated GetNimParser to use NormalizeLegacyManifest first. Created a helper ParseModelManifest to select the schema-appropriate NIM parser and parse the manifest. Legacy wrapped manifests (e.g. MolMIM) are normalized by the v1 parser. This will be invoked by the controllers.

**TESTING**: The normalizer tests in _legacy_test.go_ check detection and conversion boundaries. One case confirms a MolMIM-style fixture is recognized as legacy and produces non-empty normalized YAML. Two passthrough cases confirm normal v1 and v2 fixtures are left alone (isLegacy == false). An invalid case confirms that a legacy-looking document with a non-list manifest is flagged as legacy and returns an error. The v1 parser tests in _legacy_manifest_test.go_ check end-to-end parse behavior. _TestParseLegacyMolMIMManifest_ runs the id-list fixture through ParseModelManifestFromRawOutput, verifies two profiles, model name MolMIM, tags, and that MatchProfiles can select the TensorRT/A100 profile. _TestParseLegacySingleKeyManifest_ does the same for the single-key hash form and checks the profile ID and model name.

**Issue 2**: MoIMIM 1.0.0 will require some changes to the NimCache pathway. Similar to our standard (unset) path, MolMIM expects model_manifest.yaml and has profile selection. However, it does not have download-to-cache. Additionally, when we trigger start_server() to download the weights by providing an NGC token, the NIM image will not exit after download, but will proceed to serve and block. To download and exit, we would need to explicitly run: `python3 -c "from nimlib import nimutils; nimutils.download_models()"`

**api/apps/v1alpha1/nimcache_types.go**: _DownloadMethod_ stores what download mechanism we use
```
type NIMCacheStatus struct {....DownloadMethod string    `json:"downloadMethod,omitempty"`}
NIMCacheDownloadMethodCLI = "download-to-cache"
NIMCacheDownloadMethodNIMLib = "nimlib"
```

**internal/controller/nimcache_controller.go**: Update _getCommand()_ to check whether download-to-cache or nimlib-download is present when probing for model manifest. Create function _resolveDownloadMethod()_ to use download-to-cache when available and otherwise use nimlib. Create function _nimlibDownloadCommand()_ which will run the nimutils.downloadModels cmd: `const downloadPy = "from nimlib import nimutils; nimutils.download_models()"`. Then update to choose the correct job command in _constructJob()_:
```
if nimCache.UsesNimlibDownload() {
			command = nimlibDownloadCommand(selectedProfiles)
		} else if slices.Contains(selectedProfiles, AllProfiles) {
			args = append(args, "--all")
		} else {
			args = append(args, "--profiles")
			args = append(args, selectedProfiles...)
		}
```

**TESTING**:
_nimcache_controller_test.go_ now verifies that the probe contains both capability checks, MolMIM produces a nimlib Job and that the standard CLI is absent from that Job. _NIM_CACHE_PATH_=/model-store remains set and the all branch calls nimlib directly. Additionally, _nimcache_probe_unit_test.go_ verifies that the markers are parsed and removed from YAML, standard CLI wins if both are available, nimlib is the fallback, and that selected profiles produce _NIM_MANIFEST_PROFILE_. Additionally, both issues were verified using a MolMIM 1.0.0 image and it was confirmed that it could deploy without any issues.

### DIFF
--- a/api/apps/v1alpha1/nimcache_types.go
+++ b/api/apps/v1alpha1/nimcache_types.go
@@ -169,11 +169,17 @@ type NIMCacheStorage struct {
 
 // NIMCacheStatus defines the observed state of NIMCache.
 type NIMCacheStatus struct {
-	State      string             `json:"state,omitempty"`
-	Type       string             `json:"type,omitempty"`
-	PVC        string             `json:"pvc,omitempty"`
-	Profiles   []NIMProfile       `json:"profiles,omitempty"`
-	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+	State string `json:"state,omitempty"`
+	Type  string `json:"type,omitempty"`
+	// DownloadMethod records how model weights should be cached for this NIM
+	// image, as discovered by the manifest probe pod. Empty defaults to the
+	// standard download-to-cache CLI. "nimlib" is used for BioNeMo-style NIMs
+	// (e.g. MolMIM) that expose nimutils.download_models() but not download-to-cache.
+	// +optional
+	DownloadMethod string             `json:"downloadMethod,omitempty"`
+	PVC            string             `json:"pvc,omitempty"`
+	Profiles       []NIMProfile       `json:"profiles,omitempty"`
+	Conditions     []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
 }
 
 // NIMProfile defines the profiles that were cached.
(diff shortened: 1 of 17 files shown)


visheshtanksale (commented):
Lets keep this on hold for now. The NIM is no longer supported by Nvidia. It was last released in Aug 2024. https://catalog.ngc.nvidia.com/orgs/nim/nvidia/containers/molmim/-

We should not change NIM Operator behavior for a non supported NIM

visheshtanksale:
Closing this PR. Based on the above comment

PR #921, opened 2026-08-24
Remove NGC API Key reqs

JunAr7112:
This change is aimed at removing general NGC API Key gating requirements for the NIM Operator. Its a follow up on this earlier [issue](https://github.com/NVIDIA/k8s-nim-operator/pull/910) which removed NIMService AuthSecret Reqs for air gapped use.

**/api/apps/v1alpha1/nimcache_types.go** - set authSecret to be optional and for EnvFromSecrets() for SecretEnvSource, set Optional: &optional. If authSecret is set but the Secret or NGC_API_KEY key is missing, kubelet does not fail.

**api/apps/v1alpha1/nimservice_types.go** - Only inject NGC_API_KEY when AuthSecret is set. The secret key is optional so certain NIMs can start without it.

**internal/controller/nimbuild_controller.go** - For constructEngineBuildPod, remove the NGC_API_KEY env that used nimCache.Spec.Source.NGC.AuthSecret (would panic or point at empty name if cache had no secret). After the pod spec is built: if NGC source exists and AuthSecret != "", append NGC_API_KEY with Optional: true. Then merge user nimBuild.Spec.Env as before.

**TESTING**
nimbuild_controller_test.go, kserve/nimservice_test.go, standalone/nimservice_test.go  — switched to expect SecretKeyRef.Optional == true when the key is injected. Also verified with by deploying the NIM Operator with OSS NIM

### DIFF
--- a/api/apps/v1alpha1/nimcache_types.go
+++ b/api/apps/v1alpha1/nimcache_types.go
@@ -116,8 +116,9 @@ type HuggingFaceHubSource struct {
 // NGCSource references a model stored on NVIDIA NGC.
 // +kubebuilder:validation:XValidation:rule="!(has(self.model) && has(self.modelEndpoint))",message="Only one of 'model' or 'modelEndpoint' can be specified"
 type NGCSource struct {
-	// The name of an existing pull secret containing the NGC_API_KEY
-	AuthSecret string `json:"authSecret"`
+	// AuthSecret is the name of an existing secret containing the NGC_API_KEY.
+	// Optional for certain NIMs.
+	AuthSecret string `json:"authSecret,omitempty"`
 	// ModelPuller is the container image that can pull the model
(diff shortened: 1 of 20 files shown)


internal/controller/platform/kserve/kserve.go
```diff
@@ -63,7 +64,7 @@ func (s *KServe) Sync(ctx context.Context, r shared.Reconciler, resource client.
 		if err != nil {
 			if errors.IsConflict(err) {
 				// Ignore conflict errors and retry.
-				return ctrl.Result{Requeue: true}, nil
+				return ctrl.Result{RequeueAfter: time.Second}, nil
```
visheshtanksale, line 67: Is this change required?

visheshtanksale (commented):
Minor comment. Looks good overall

internal/controller/platform/kserve/kserve.go
JunAr7112: No this change isn't needed. Resolved by rebasing to main

visheshtanksale (approved):

